PR #12, opened 2024-02-23
Added support for binary, octadecimal and hexadecimal numbers

JJtan2002:


src/tokenizer.ts
```diff
@@ -196,11 +196,54 @@ export class Tokenizer {
     private isDigit(c: string): boolean {
         return /^[0-9]/.test(c);
     }
-
+	
+	private isHexa(c: string) : boolean {
```
Fidget-Spinner, line 200: What IDE are you using? Could you please set it to use spaces instead of tabs for indentation please? The rest of the file uses 4 spaces for an indent.

Please fix the indentation for the rest of the file too.

src/tokenizer.ts
```diff
@@ -196,11 +196,54 @@ export class Tokenizer {
...
+		return /^[0-1]/.test(c);
+	}
+	
     private isIdentifier(c: string): boolean {
         return c === '_' || this.isAlpha(c) || this.isDigit(c);
     }
-
+	
+	private baseNumber() {
+		if (this.peek() === 'x') {
```
Fidget-Spinner, line 217: Please change this to a `switch-case` statement. It's clearer. So it should be
```
switch(this.peek()) {
    case 'x': ...
    case 'o': ...
}
```

Fidget-Spinner (commented):
Good work! A few comments from me.

Fidget-Spinner (approved):

PR #15, opened 2024-03-11
Issue14

JJtan2002:


Fidget-Spinner:
Thanks for the quick response! Could you please paste a screenshot here to check how the error message looks like?

JJtan2002:
image
The error is thrown when there are no trailing digits after e, x, b, o. Otherwise the expression would unexpectedly evaluate to null.

Fidget-Spinner (approved):

Issue #8, opened 2023-05-08
Parser/tokenizer does not support triple quoted strings


Fidget-Spinner:
Triple quoted strings should be supported. An example of their usage is docstrings.

```python
def foo():
    """Some docs here"""
```

PR #16, opened 2024-03-15
Fix issue #8

JJtan2002:
Added support for triple-quoted multi-line strings with double and single quotes.
Fixed a bug where statements consisting of a single string gives syntax error unexpectedly.

### DIFF
--- a/src/tokenizer.ts
+++ b/src/tokenizer.ts
@@ -177,14 +177,14 @@ export class Tokenizer {
         const lexeme = this.source.slice(this.start, this.current);
         this.tokens.push(new Token(type, lexeme, line, col, this.current - lexeme.length))
     }
-
+    // Ensures that lexeme value is parsed correctly by removing starting and ending quotes.
     private addStringToken(type: TokenType) {
         const line = this.line
         const col = this.col;
         const lexeme = this.source.slice(this.start + 1, this.current - 1);
         this.tokens.push(new Token(type, lexeme, line, col, this.current - lexeme.length))
     }
-
+    // Remove three starting and ending quotes
     private addMultiLineStringToken(type: TokenType) {
         const line = this.line
         const col = this.col;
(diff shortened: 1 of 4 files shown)


src/tokenizer.ts
```diff
@@ -182,6 +185,12 @@ export class Tokenizer {
         this.tokens.push(new Token(type, lexeme, line, col, this.current - lexeme.length))
     }
 
+    private addMultiLineStringToken(type: TokenType) {
+        const line = this.line
+        const col = this.col;
+        const lexeme = this.source.slice(this.start + 3, this.current - 3);
```
Fidget-Spinner, line 191: Might want to add a comment explaining why +3 and -3 here. I understand that it's due to the `"""` but that's not always obvious to others :).

src/tokenizer.ts
```diff
@@ -432,26 +441,47 @@ export class Tokenizer {
...
+                                this.col, this.source, this.start, this.current);
+                        }
+                    }
+                } else {
+                    while (this.peek() != quote && this.peek() != '\n' && !this.isAtEnd()) {
+                        this.advance();
+                    }
+                    if (this.peek() === '\n' || this.isAtEnd()) {
+                        throw new TokenizerErrors.UnterminatedStringError(this.line, this.col, this.source, this.start, this.current);
+                    }
```
Fidget-Spinner, line 480: So for this section, could you please use a different style of checks?

The idea is that we don't want to form highly-nested code (you'll learn more about this in CS2103T), so instead of
```
if (cond) {
} else {
}
```
Do
```
if (cond) {
# return early/ fail early
}
# Rest of the code
```

E.g.

```
                        if (this.peek() == quote) {
                            this.advance(); // consume second ending quote
                            if (this.peek() == quote) {
                                this.advance(); // consume third ending quote
                                this.addMultiLineStringToken(TokenType.STRING);
                            } else {
                                throw new TokenizerErrors.UnterminatedStringError(this.line,
                                    this.col, this.source, this.start, this.current);
                            }
                        } else {
                            throw new TokenizerErrors.UnterminatedStringError(this.line,
                                this.col, this.source, this.start, this.current);
                        }
```
should be (notice how much more linear and less nested the code is!)
```
                        if (this.peek() != quote) {
                            throw new TokenizerErrors.UnterminatedStringError(this.line,
                                this.col, this.source, this.start, this.current);
                        }
                        this.advance(); // consume second ending quote
                        if (this.peek() != quote) {
                            throw new TokenizerErrors.UnterminatedStringError(this.line,
                                this.col, this.source, this.start, this.current);
                        }
		        this.advance(); // consume third ending quote
		        this.addMultiLineStringToken(TokenType.STRING);

```
please apply this to the whole section.

src/tokenizer.ts
```diff
@@ -432,26 +441,48 @@ export class Tokenizer {
...
-            case '\'':
-                while (this.peek() != '\'' && this.peek() != '\n' && !this.isAtEnd()) {
+            case "'":
+                let quote = c;
+                if (this.peek() == quote) { // handle multi-line string
+                    this.advance(); // second quote found and consumed
+                    if (this.peek() != quote) { // empty string ""
+                        this.addStringToken(TokenType.STRING);
+                        break;
+                    } else {
```
Fidget-Spinner, line 451: No need else block here, can just flatten and bring out this.advance() because you already put a break in the if block

src/tokenizer.ts
```diff
@@ -432,26 +441,48 @@ export class Tokenizer {
...
+                    }
+                    while (this.peek() != quote && !this.isAtEnd()) {
+                        this.advance(); // advance until ending quote found
+                    }
+                    if (this.isAtEnd()) {
+                        throw new TokenizerErrors.UnterminatedStringError(this.line,
+                            this.col, this.source, this.start, this.current);
+                    }
+                    this.advance(); // consume first ending quote
+                    if (this.peek() == quote) {
```
Fidget-Spinner, line 462: You can invert the logic to flatten it here. So check for this.peek() != quote, if so error.

A lot of professional code has the style:


```
if (unexpected) {
    error
}
if (other unexpected) {
    error
}
happy case
```

src/tokenizer.ts
```diff
@@ -432,26 +441,48 @@ export class Tokenizer {
...
+                        this.advance(); // consume second ending quote
+                    } else {
+                        throw new TokenizerErrors.UnterminatedStringError(this.line,
+                            this.col, this.source, this.start, this.current);
+                    }
+                    if (this.peek() == quote) {
```
Fidget-Spinner, line 468: Same as above comment about flattening by inverting logic

src/tokenizer.ts
```diff
@@ -432,26 +441,48 @@ export class Tokenizer {
...
+                        this.advance(); // consume third ending quote
+                        this.addMultiLineStringToken(TokenType.STRING);
+                    } else {
+                        throw new TokenizerErrors.UnterminatedStringError(this.line,
+                            this.col, this.source, this.start, this.current);
+                    }
+                } else { // other case, single-line string
```
Fidget-Spinner, line 475: You can remove the else here by adding a break in the previous line to end the if. That removes one layer of nesting.

src/tokenizer.ts
```diff
@@ -174,14 +177,20 @@ export class Tokenizer {
         const lexeme = this.source.slice(this.start, this.current);
         this.tokens.push(new Token(type, lexeme, line, col, this.current - lexeme.length))
     }
-
+    // Ensures that lexeme value is parsed correctly by removing starting and ending quotes.
```
Fidget-Spinner, line 180: Comments should be as close to the source as they are describing. In this case it should be placed on a separate line above the this.source.slice

src/tokenizer.ts
```diff
@@ -174,14 +177,20 @@ export class Tokenizer {
...
     private addStringToken(type: TokenType) {
         const line = this.line
         const col = this.col;
         const lexeme = this.source.slice(this.start + 1, this.current - 1);
         this.tokens.push(new Token(type, lexeme, line, col, this.current - lexeme.length))
     }
-
+    // Remove three starting and ending quotes
```
Fidget-Spinner, line 187: Same here, please move to a new line above this.source.slice

Fidget-Spinner (approved):
Ok I hope you tested it. You can merge it now. Thanks for addressing my reviews!

PS.
* Please mark a comment as resolve once you address them.
* When merging, make sure to give a descriptive title. Maybe something like "Support triple quoted multiline strings"

PR #49, opened 2025-04-03
Feature translator

WangYuyang1013:
**Project Environment**
In package.json, new dependencies required for Rollup bundling as well as other newly used packages have been added. The build logic has been switched from tsc to Rollup to better support integration with the Source Academy frontend. A new Rollup configuration file was created to bundle src/index.ts into dist/index.js, making it easier to deploy and use.
In tsconfig.json, the module type was changed from commonjs to ESNext to support bundling, and the exclude list was updated to ignore directories and files such as dist, docs, and tests.

**Codebase Changes**
In the newly added file types.ts, two AST node types None and ComplexLiteral were introduced to represent Python’s None and complex numbers respectively. The class PyComplexNumber was also introduced to store and compute complex numbers. Additionally, some error-handling logic was added to the file.

In translator.ts, support was added for translating the new None and ComplexLiteral nodes.

In ast-types.ts, two new expression nodes were added: None and Complex. The Complex node stores its value using the PyComplexNumber class after parsing a complex number string during construction.

For tokenizer.ts
Enhanced string handling logic: Supports parsing multi-line strings enclosed in triple quotes, and accurately handles escape characters such as \n, \\, and \". Use lexemeBuffer to record the raw content of strings. Supports detection and handling of illegal escape characters (e.g., \z will throw a SyntaxWarning).
Identifiers can now include legal Unicode characters, validated using the isLegalUnicode function.
Complex numbers: Recognizes characters j and J in numeric contexts as indicating imaginary numbers.
Underscore in numeric literals: Numbers like 1_000_000 are now parsed correctly.

In tokens.ts:
A new token type COMPLEX was added to represent complex number literals.

In parser.ts:
Parsing logic was enhanced to handle integers with underscores and complex numbers.

In resolver.ts:
Built-in Python functions are now handled by registering their names into the environment (without handling function logic yet). Handling logic for the None and ComplexLiteral AST nodes was also added.


src/types.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    private toPythonComplexFloat(num: number){
+        if (num === Infinity) {
+            return "inf";
+        }
+        if (num === -Infinity) {
+            return "-inf";
+        }
+        
+        if (Math.abs(num) >= 1e16 || (num !== 0 && Math.abs(num) < 1e-4)) {
```
Fidget-Spinner, line 214: Can you add a comment stating how you derived these bounds? These magic numbers are a little confusing otherwise.

Fidget-Spinner (commented):
Really great work! Just have one comment.

src/types.ts
WangYuyang1013: Really appreciate that! I'm trying to fully replicate the behavior of the Python interpreter, but there are still some precision differences in the csemachine part, which will be coverred later on.

Fidget-Spinner (approved):
OK let's merge it after you add the comment. I'll leave it to you to merge.

WangYuyang1013:
Thanks for your review! I've added the comment and just merged the PR as discussed.

PR #53, opened 2025-04-28
some updates on resolver

WangYuyang1013:
Sorry for the delay in submitting this PR—I was quite busy over the past two weeks and couldn't get to it earlier.

During last week's testing, I discovered an issue. In cases where a function is defined first and then a variable with the same name is declared, the code does not throw a redefinition error. Another similar issue is that the function environment was not exited after the function ended. example:

```x = 1
def a():
    return 1
x = 2
```

This would not raise any errors.

After making the following changes, these cases now behave as expected:
Added a new property definedNames and updated the declareName method:
This property keeps track of all names that have been defined in the current environment, preventing redefinitions—particularly guarding against redeclarations involving RedefineableTokenSentinel.
In visitFunctionDefStmt:
I uncommented the line that saves the current environment (oldEnv) and function scope so that after resolving the function body, the environment correctly restores to the previous one.

May I ask, what was the original intention behind commenting out the oldEnv saving logic in visitFunctionDefStmt? Was there a specific consideration for doing so?

### DIFF
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -21,12 +21,14 @@ class Environment {
     // This only set at the top level environment. Child environments do not
     // copy this field.
     moduleBindings: Set<string>;
+    definedNames: Set<string>;
     constructor(source: string, enclosing: Environment | null, names: Map<string, Token>) {
         this.source = source;
         this.enclosing = enclosing;
         this.names = names;
         this.functions = new Set();
         this.moduleBindings = new Set();
+        this.definedNames = new Set();
     }
 
     /*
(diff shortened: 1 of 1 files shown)


Fidget-Spinner:
> May I ask, what was the original intention behind commenting out the oldEnv saving logic in visitFunctionDefStmt? Was there a specific consideration for doing so?

Don't remember. It might have been buggy at the time?

Fidget-Spinner (approved):

WangYuyang1013:
> > May I ask, what was the original intention behind commenting out the oldEnv saving logic in visitFunctionDefStmt? Was there a specific consideration for doing so?
> 
> Don't remember. It might have been buggy at the time?

I also think it might have been buggy at the time, but after testing now, I haven't encountered any issues. Let's keep it as it is for now.

PR #66, opened 2025-09-20
Import Conductor as package

loyaltypollution:
Previously, py-slang copied the Conductor framework code directly into the repo. This PR replaces that approach by consuming Conductor as an npm library instead.

I worked with @tsammeow to expose the correct library API for py-slang to import from.
This change significantly reduces the codebase size of py-slang while keeping functionality the same.

- Removes duplicated Conductor code in py-slang
- Centralizes Conductor maintenance in its own package
- Decreases codebase size


s-kybound:
I really like the direction this PR is taking with the conductor - it should be standardized as a package instead of being copied across all of the evaluators. will take a closer look the PR.

martin-henz:
Note that we are trying to adhere to these new conventions for Source Academy projects: https://github.com/source-academy/general/wiki/Conventions-and-practices

If there is a need for discussion of those, let's use issues in the repo "general".

WangYuyang1013 (approved):
This PR takes a very good direction, and I appreciate the effort to reduce duplication and standardize the use of Conductor as a package. While I may not fully understand all the code changes in detail, I trust that if it works correctly on your side, it should be fine.

s-kybound (changes requested):
again, a resounding `yes` for this direction. but is conductor not meant to be imported as a NPM library `@sourceacademy/conductor`, and not a git repo?

cc @tsammeow 

once resolved, LGTM

martin-henz:
> I'm looking at frontend and other repos like this, it seems like we are currently using tagged github repositories for all package.json.
> 
> Is it because our work is still in development? @tsammeow @s-kybound

Yes, let's use NPM for such dependencies. We have a few packages in NPM with prefix sourceacademy already: https://www.npmjs.com/search?q=sourceacademy
...and of course js-slang is an NPM package as well:
https://www.npmjs.com/package/js-slang

Let's follow our guide on this:

https://github.com/source-academy/general/wiki/Conventions-and-practices

PR #81, opened 2026-03-03
Python chapter 2 (+ 3?) support

AaravMalani:
So far (updating it as more gets pushed to the branch), the PR 
- Introduces chapter-based evaluators, so every chapter generates a different `worker.js`
- Improves the type safety of `Value` types
- Load groups (`LINKED_LIST`, `LIST`, `PAIRMUTATORS`, `STREAMS`, etc.) as a combination of native functions and preludes
- Create preludes and native functions for the aforementioned groups (except certain functions such as `draw_data`)
- Create a testing framework and implement it for `stdlib` and `linked-list`
- Fix rounding to use banker's rounding instead of rounding to the nearest integer
- Add some builtin functions (`is_none`, `is_function`, `is_float`, `is_string`, `is_boolean`)
- Added the list literal creation instruction to the CSE machine and restricted parsing of arrays to chapter 3 onwards
- Add list indexing and assignment
- Add while loops
- Merged `undefined` into `None`

Note: ignore files which say they're not in use, we're figuring out how/whether to merge them into the current code

### DIFF
--- a/docs/lib/linked_list.py
+++ b/docs/lib/linked_list.py
@@ -125,10 +125,8 @@ def equal(xs, ys):
         )
     elif is_none(xs):
         return is_none(ys)
-    elif is_int(xs):
-        return is_int(ys) and xs == ys
-    elif is_float(xs):
-        return is_float(ys) and xs == ys
+    elif is_int(xs) or is_float(xs) or is_complex(xs):
+        return (is_int(ys) or is_float(ys) or is_complex(ys)) and xs == ys
     elif is_boolean(xs):
         return is_boolean(ys) and ((xs and ys) or (not xs and not ys))
     elif is_string(xs):
--- a/docs/specs/python_typing.tex
+++ b/docs/specs/python_typing.tex
@@ -68,6 +68,6 @@ of arguments and generate an error message when the types do not match.
   \end{tabular}
 \end{center}
 
-Preceding conditional expressions and following if statements,
+Following \texttt{\textbf{if}} and \texttt{\textbf{elif}},
 Python only allows boolean expressions. Lambda expressions can only
-be used as arguments to operators when they are paranthesized.
+be used as arguments to operators when they are parenthesized.
--- a/docs/specs/python_typing_3.tex
+++ b/docs/specs/python_typing_3.tex
@@ -68,6 +68,6 @@ of arguments and generate an error message when the types do not match.
   \end{tabular}
 \end{center}
 
-Preceding \texttt{\textbf{?}} and following \texttt{\textbf{if}},
+Following \texttt{\textbf{if}} and \texttt{\textbf{elif}},
 Python only allows boolean expressions. Lambda expressions can only
-be used as arguments to operators when they are paranthesized.
+be used as arguments to operators when they are parenthesized.


jonasongg:
@AaravMalani is there any way to take the linked-list prelude from directly from linked-list.py under docs/ so we have a single source of truth for the prelude?

AaravMalani:
> @AaravMalani is there any way to take the linked-list prelude from directly from linked-list.py under docs/ so we have a single source of truth for the prelude?

I'd rather update the `equal` function in the documentation so that `equal(1, 1.0)` is `True` first (I think we discussed this earlier) before merging it in

src/tests/analysis.test.ts
s-kybound: Why are the tests here commented out? If we are using them in the future, please leave a TODO on each one so that others will know to work on them in the future.

If not, please leave some comment explaining why they are removed/remove them altogether

src/tests/resolver.test.ts
```diff
@@ -0,0 +1,186 @@
...
+def foo():
+    print(x)
+foo()
+            `;
+      expect(toPythonAstAndResolve(code, 1)).toMatchObject({});
+    });
+  });
+
+  describe("Variant Specific Syntax", () => {
+    //         test('For loops throw errors for Python 1 and 2', () => {
```
s-kybound: Why is this test commented out?

src/types.ts
```diff
@@ -271,6 +278,12 @@ export interface Finished {
   // field instead
 }
 
+// export interface CSEError {
+//     status: 'error'
+//     context: Context
+//     error: Error
+// }
+
 // export class Representation {
```
s-kybound: ditto

src/cse-machine/operators.ts
```diff
@@ -88,31 +95,99 @@ export function evaluateUnaryExpression(
...
+        operatorTranslator(operator),
+      ),
+    );
+  }
 
+  // Handle complex number equality
+  if (left.type == "complex" || right.type == "complex") {
+    if (
+      (right.type !== "complex" && right.type !== "number" && right.type !== "bigint") ||
+      (left.type !== "complex" && left.type !== "number" && left.type !== "bigint")
```
s-kybound: maybe we should create helper functions to check for different numbers?

ie `isNumeric` to check for any type of number, `isComplex` (which is == `isNumeric` ) that checks for any type of number that can be coerced into a complex number, etc... or strict variants if we only want that specific type of number

src/cse-machine/operators.ts
```diff
@@ -605,25 +683,23 @@ export function evaluateBoolExpression(
   context: Context,
   operator: TokenType,
   left: Value,
-  right: Value,
```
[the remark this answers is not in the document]
s-kybound: Actually, it seems like this is dead code. Shall we remove this function completely?

s-kybound (changes requested):
Nice progress here - just some comments on the PR to polish:
- Commented-out code - Several regions of the code commented out without explanation. if not searchable via TODO + justified via comments, future developers will probably ignore these, leading to slow code rot. Could we mark by TODO or remove entirely?
- Semantics - New changes seem to break truthiness/falsiness semantics in python, restricting some operations to boolean only. Some worries on None semantics as well. Could we clarify the semantics in these regions?
- Code quality - some parts of the code can be made more readable, and others (such as `evaluateBoolExpression`) are confusing without required context in different parts of the codebase. Might help to add a few comments/better names to clarify the intent here.

PR #82, opened 2026-03-04
Add Python Chapter 2 documentation

jonasongg:
Includes change to source_typing (`and` and `or` operators), which changes source_1 also. Also remove latex auxiliary files git tracking.

### DIFF
--- a/docs/python/lib/list.py
+++ b/docs/python/lib/list.py
@@ -130,13 +130,11 @@ def equal(xs, ys):
     elif is_float(xs):
         return is_float(ys) and xs == ys
     elif is_boolean(xs):
-        return is_boolean(ys) and xs == ys
+        return is_boolean(ys) and ((xs and ys) or (not xs and not ys))
     elif is_string(xs):
         return is_string(ys) and xs == ys
-    elif is_function(xs):
-        return is_function(ys) and xs == ys
     else:
-        return xs == ys
+        return False
 
 
 def length(xs):
--- a/docs/python/specs/source_typing_3.tex
+++ b/docs/python/specs/source_typing_3.tex
@@ -45,10 +45,8 @@ operator & argument 1 & argument 2 & result\\ \hline
 \texttt{\textbf{\%}} & int   & int   & int\\
 \texttt{\textbf{\%}} & int   & float   & float\\
 \texttt{\textbf{\%}} & float   & float   & float\\
-\texttt{\textbf{==}} & int,float,complex   & int,float,complex     & bool\\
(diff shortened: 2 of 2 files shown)


martin-henz (commented):
Remove all generated files: pdf
from the PR. Version control should not include files that are derived from other files. The input in this case are the .tex files

docs/python/lib/list.py
```diff
@@ -0,0 +1,367 @@
...
+            is_pair(ys)
+            and equal(head(xs), head(ys))
+            and equal(tail(xs), tail(ys))
+        )
+    elif is_none(xs):
+        return is_none(ys)
+    elif is_number(xs):
+        return is_number(ys) and xs == ys
+    elif is_boolean(xs):
+        return is_boolean(ys) and xs == ys
```
AaravMalani: Isn't equality not defined for booleans and functions as per the spec?

docs/python/lib/list.py
```diff
@@ -0,0 +1,367 @@
...
+    )
+
+
+def list_ref(xs, n):
+    """
+    Returns the element of linked list xs at position n (0-indexed).
+    """
+    if n == 0:
+        if is_none(xs):
+            raise IndexError(
```
AaravMalani: Although this is probably pseudocode to some extent, I think using `error` would make more sense so the output is more predictable to someone reading the specs since you can't `raise` in the py-slang

docs/python/lib/list.py
```diff
@@ -0,0 +1,367 @@
...
+    Returns:
+        value: head of p
+    """
+    pass
+
+
+def tail(p):
+    """
+    PRIMITIVE
+    Returns tail (second component of given pair p.
```
AaravMalani:
```suggestion
    Returns tail (second component) of given pair p.
```

docs/python/lib/list.py
AaravMalani: Is `print_linked_list` (adapted from `display_list`) going to be part of the list functions?

docs/python/lib/list.py
jonasongg: actually not entirely sure if we should raise at all, since this doesn't exist in Source. we might assume that the input is valid. will check with prof martin on this

jonasongg:
@AaravMalani thanks for the comments and great catches!

docs/python/specs/source_2_bnf.tex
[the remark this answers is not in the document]
jonasongg: no, i don't think so

Issue #11, opened 2023-06-24
Support Full Python using Pyodide


RichDom2185:
[Pyodide](https://pyodide.org/en/stable/) is a port of CPython to WebAssembly.

I was discussing with @Fidget-Spinner regarding the possibility of adding support for Full Python into Source Academy by way of using Pyodide. The following was a summary:

* Pros:
  * Students can experiment with the full Python syntax online, without having to install Python locally
  * Python modules such as NumPy work and can be imported
* Cons:
  * Our own Source Academy modules (e.g. Runes) won't be supported
  * Our JavaScript bundle size will be quite bloated (Pyodide alone is 6.4MB)

@martin-henz what do you think of this potential enhancement?

Fidget-Spinner:
After more digging around, I realised that we can indeed import JS modules from Pyodide. This would require some rework for our modules and i would ezpect it to be roughly equivalent to a CP3108B project.

https://pyodide.org/en/stable/usage/api/python-api/ffi.html#pyodide.ffi.register_js_module

The only drawback is that our current tools wouldnt be able to visualize full Python. However, that doesnt matter anyways was full JS also doesnt support env viz, stepper, etc.

So in short we could have:

Python 1-4: py-slang
Full Python: Pyodide + import from js-slang

RichDom2185:
That looks promising! Pyodide also has a built-in React component for an interactive console/REPL, and it seems CS1010S is using it too:

https://github.com/NUSSOC/glide

martin-henz:
Yes, this sounds promising. It would be great to support full Python via Pyodide. Then Python would be on par with JavaScript in Source Academy (modulo SICP...).

loyaltypollution:
Just to revive an old issue, I have implemented pyodide in conductor framework here: https://github.com/loyaltypollution/py-slang-pyscript

Should we keep as separate slang or integrate it?

martin-henz:
I think the same slang can have multiple evaluators.

So for Python, let's aim for a single repository py-slang. The build process should build separate bundles and deploy them in a public place, usually a github page.

Let's discuss this in the "languages" telegram group.

PR #117, opened 2026-03-26
Add Pyodide implementation

veehz:


martin-henz:
Let's look at this after the Python migration. It's not mission-critical right now.

martin-henz:
Superseding this with #312 — a fresh port against the current architecture (the evaluator naming, build system, and conductor version have all moved on significantly since this was opened in March). #312 covers the plain Pyodide evaluator (this PR's, and #11's, original ask); the PyTorch bridge work here is tracked separately in #313 so it isn't blocking the base evaluator.

Thanks for the original implementation, @veehz — the core approach (parse+analyze through py-slang's own validator before handing off to pyodide) carried over directly, and `bridge.py`/`loadTorch.ts`/`@sourceacademy/torch` are all still right here in this branch's history for #313 whenever someone picks it up.

PR #124, opened 2026-03-29
Fix types of Python grammar

AaravMalani:
Right now, the Python grammar generated by Nearley has several Typescript errors. These include
- Missing variable names behind conditional checks that never return false (`pythonLexer.has('...') ? ... : missingVariable`). The Moo lexer's `has` function always returns `true`, so the second branch is never executed leading to dead code which still throws errors
- Implicit `any`s / missing types: All the post-processing functions lack types which may not be required, but they still throw errors and lack clarity (e.g., the recent if/else bug would have been easier to detect with proper typing0

This PR,
- Removes the redundant ternaries as mentioned above
- Adds proper typing to the grammar

_Note: The method of post-processing the file with `sed` may not be the most maintainable, but this issue has been documented upstream in [Nearley](https://github.com/kach/nearley/issues/576) and the current method seems like the best method till Nearley supports generating Typescript grammars)_


AaravMalani:
Great, thanks!

PR #131, opened 2026-03-30
Restructure directories

loyaltypollution:
We have an increasing number of evaluators for Python:

- cse machine
- wasm compiler
- svml interpreter (upcoming PR)
- pyodide backend

Commit e908db8d6253ce578c85e41b268074231f8d01d1 tries to handle this by splitting into evaluators with inheritance patterns.

Let's use a better restructure altogether


src/index.ts
```diff
@@ -0,0 +1,17 @@
+declare const __EVALUATOR__: string;
+
+import { initialise } from "@sourceacademy/conductor/runner";
+
+import { PyCSEEvaluator } from "./conductor/PyCSEEvaluator";
+import { PyWasmEvaluator } from "./conductor/PyWasmEvaluator";
```
RichDom2185: I just realised this is inconsistent. Should be `PyCseEvaluator` no? Since JS treats abbreviations and acronyms as a single word

RichDom2185 (approved):
Just the one comment, otherwise IPA LGTM contingent on others' reviews

scripts/build.ts
```diff
@@ -0,0 +1,58 @@
...
+  cse: "PyCSEEvaluator",
+  wasm: "PyWasmEvaluator",
+} as const;
+
+type EvaluatorKey = keyof typeof evaluatorMap;
+const evaluators = Object.keys(evaluatorMap) as EvaluatorKey[];
+
+function buildEvaluator(name: EvaluatorKey) {
+  console.log(`\nBuilding evaluator=${name} (${evaluatorMap[name]})...\n`);
+  execSync("rollup -c --bundleConfigAsCjs", {
```
AaravMalani: The issue I have wrapping the `rollup` CLI is that I can't pass additional flags to the `rollup` binary
I usually run `rollup` with the `--watch` flag which speeds up incremental builds by a ton

Is there a way to add this?

src/conductor/PyCSEEvaluator.ts
```diff
@@ -1,69 +1,91 @@
...
+        groups: ALL_GROUPS,
+      });
+
+      if (result.type === "error") {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        this.conductor.sendError(toEvaluatorError(new Error(result.message)) as any);
+        return;
+      }
+
+      this.conductor.sendResult(cseValueToNative(result));
```
AaravMalani: Is there a reason we're sending the result? Python isn't supposed to do that as per the spec

src/conductor/PyCSEEvaluator.ts
```diff
@@ -1,69 +1,91 @@
...
+      const ast = parse(script);
+      const errors = analyze(ast, script, 4, ALL_GROUPS);
+      if (errors.length > 0) {
+        for (const error of errors.slice(0, -1)) {
+          await displayError(context, error, ErrorType.EVALUATOR_SYNTAX);
+        }
+        throw errors[errors.length - 1];
+      }
+      const result = await evaluate("", ast, context, {
+        variant: 4,
```
AaravMalani: Is there a reason for hardcoding Python 4?

package.json
```diff
@@ -18,7 +20,7 @@
     "lint": "eslint --concurrency=auto src",
     "format": "prettier --write \"**/*.{ts,tsx,json,js,mjs}\"",
     "format:ci": "prettier --list-different \"**/*.{ts,tsx,json,js,mjs}\"",
-    "wasm": "ts-node --transpile-only src/wasm-compiler/run-file.ts"
+    "wasm": "ts-node --transpile-only src/engines/wasm/run-file.ts"
```
AaravMalani: Maybe we should move over to `tsx` for consistency then

src/index.ts
AaravMalani: Agreed, think this is on me when I made the change (meant to reply to Richard's comment)

src/conductor/PyCSEEvaluator.ts
```diff
@@ -1,69 +1,85 @@
...
+    case "none":
+      return undefined;
+    default:
+      return undefined;
   }
+}
 
+export class PyCSEEvaluator extends BasicEvaluator {
   async evaluateChunk(chunk: string): Promise<void> {
+    const context = new Context();
```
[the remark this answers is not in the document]
AaravMalani: Is there a reason this was resolved? This seems like a valid issue

src/conductor/errors.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Map an engine-thrown exception to an error suitable for conductor.sendError().
+ *
+ * In production, the conductor accepts ConductorError subclasses. However, the
+ * @sourceacademy/conductor package's minified ESM error classes break under
+ * Jest's CJS transform (class field initializers before super()). We use a
+ * plain object that satisfies the structural interface instead.
+ */
+export function toEvaluatorError(e: unknown): { message: string; line?: number; column?: number } {
```
AaravMalani: The conductor error requires the `name` and `errorType` field, adding those may remove an unnecessary `as any` in the `PyCSEEvaluator.ts` file
Super minor thing though

AaravMalani (changes requested):
Most stuff is minor, but is there a reason Python 4 is hard-coded (just wanted to know this before I approve)?
Also, this may be on my setup, but I get a ton of `Us[n] is not a function` errors when I run code (I had something similar when using Terser, could be something on my end)

Thanks for the PR!

package.json
AaravMalani, line 1: Should tsx be a dev dependency? We use it a lot, might make sense over `npx`ing it for every script

RichDom2185: Agreed, let's make it a dev dep

package.json
jonasongg: i think if we do this, we can also just remove ts-node as a dev dependency. right now it's used for start:dev > nodemon, but that doesn't do anything anyway now that src/index is empty @loyaltypollution

rollup.config.mjs
```diff
@@ -1,87 +1,63 @@
...
+  },
 });
 
-const variants = [1, 2, 3].map(v => generateConfig(v));
+function plugins() {
+  return [
+    replacePlugin,
+    commonjs({ include: "node_modules/**" }),
+    json(),
+    wasm({ maxFileSize: 100_000 }),
```
jonasongg: @loyaltypollution is there a reason this plugin is added? as far as i know, it's not needed

yytelliot:
Does the evaluator update its status? I don't seem to be able to find any updates to evaluation status in the code. I think this might have to be addressed in a future PR. See https://github.com/source-academy/conductor?tab=readme-ov-file#updating-status

martin-henz:
> Does the evaluator update its status? I don't seem to be able to find any updates to evaluation status in the code. I think this might have to be addressed in a future PR. See https://github.com/source-academy/conductor?tab=readme-ov-file#updating-status

Yes, let's complete the restructuring first, and then look at the status updates for each evaluator in separate PRs.

rollup.config.mjs
loyaltypollution: in upcoming PR, I'll use it to run sinter in WASM port, I've split this work from the older PR

src/conductor/PyCSEEvaluator.ts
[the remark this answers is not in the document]
loyaltypollution: agreed, I was trying to bug catch a separate issue:

if we have multiple evaluator instances, they will both share and conflict with global level Context. I think the correct fix is to instantiate context as Evaluator level.

src/index.ts
loyaltypollution: agreed, changed

scripts/build.ts
loyaltypollution: I could try piping argon into the rollup object, or add a convenience command like build:watch

src/conductor/PyCSEEvaluator.ts
loyaltypollution: bug on my part, I was mapping out PyCseEvaluator 4, let's modify the build script to also handle variants

package.json
loyaltypollution: agreed, easy fix of internal packages

scripts/build.ts
loyaltypollution: nevermind, its likely easier to expand build.ts to support --watch directly

AaravMalani:
Another issue I'm noticing is that because we've eagerly imported the classes in the `index.ts` file, it isn't tree-shaking, and every evaluator is being included in the final bundle. I think moving the initialisations back into their own file would fix that (`PyCseEvaluator1.ts`, etc.), or some other way.

AaravMalani:
> Another issue I'm noticing is that because we've eagerly imported the classes in the `index.ts` file, it isn't tree-shaking, and every evaluator is being included in the final bundle. I think moving the initialisations back into their own file would fix that (`PyCseEvaluator1.ts`, etc.), or some other way.

Just waiting for a fix to this, otherwise LGTM

loyaltypollution:
Fixed tree shaking issue @AaravMalani 

seems like being precise with rollup config is sufficient: for now ive declared moduleSideEffects = false

this assumes no side effect from module imports. if there are exceptions (like polyfills), rollup lets us notate exceptions to moduleSideEffect accordingly

AaravMalani (approved):
Works, LGTM!
Thanks for the PR

PR #145, opened 2026-04-15
General Fixes

AaravMalani:
- Refactor some code and delete some unneeded files
- Fix preludes not being resolved properly with the Conductor evaluators
- Fixed the list functions not working with the Conductor evaluators
- Replace every use of `constructor.name` so we can mangle function names with Terser
- Narrow the types in command evaluators to prevent unnecessary `as` casts
- Fix the `StepLimitExceededError` exception
- Fix `==` in expanded Python 3 equality
- Add test cases for Python 2 and 3 standard libraries


martin-henz (approved):
Substantial improvements! Thanks!

PR #149, opened 2026-04-20
Wasm compiler conductor integration and cleanup

jonasongg:
* Adds the 4 Wasm evaluators to build.
* Split up the giant `constants.ts` into the `runtime` folder
* Split up the giant `wasm-compiler.spec.ts` into the `wasm` folder
* Split up `index.ts`
* **Ban for loop without the range syntax even for Chapter 4** (modified analysis.ts)
* **Remove BreakContinueError** as it's not used (the validation is done elsewhere, and the error is defined elsewhere; see `BreakContinueOutsideLoopError`

Work that still needs to be done for wasm compiler:
* Finish up misc & math libraries, which are not done (especially `repr` function: it's wrong at the moment)
* Potentially move certain validations out of `builderGenerator` into validators, since they should apply to all python implementations
  * Perhaps give a more specific error if a name is used before its nonlocal declaration in the same scope (right now, validator gives generic NameNotFoundError)
  * Give a more specific error if parameters are used as nonlocal also
  * Error when more than one starred parameter is used
  * Error when starred parameter is not the last one


martin-henz:
Somehow this PR didn't get reviewed and merged last semester. Let's KIV this because it is not mission-critical for our Python conversion.

Issue #110, opened 2026-03-24
Add more testing for the CSE machine and the Conductor evaluator


AaravMalani:
The parser, resolver and validator are pretty well tested. However, the CSE machine is basically untested so far. There are two aspects of testing needed to be done

1. Testing the CSE machine: This is something #81 introduces extensively for Python 1 and 2. It builds a testing framework which uses `runInContext` to evaluate programs. It still lacks support for Python 3 and 4, although that 

2. Testing Conductor-specific features: The issue with the `runInContext` method is it bypasses the Conductor entirely. Features like IO streams from #107 and Autocomplete from #109 require a Conductor instance. IMO, there probably needs to be a mock Conductor framework which is testable

martin-henz:
cse machine is now well tested

Akshay-2007-1:
> cse machine is now well tested

Prof, the mock conductor framework testing is not in place yet afaik. @AaravMalani, could you confirm? If thats the case, this issue is still OPEN.

PR #152, opened 2026-06-10
feat(stepper): implement PythonStepperEvaluator and add stepper dependencies

Shrey5132:
This PR adds portal dependencies on `@sourceacademy/common-stepper` and `@sourceacademy/runner-stepper` from the `plugins` repository. It implements `PythonStepperEvaluator1` and its supporting AST walking, translation, and state reduction logic to step through Python 1 code using the Conductor framework.


scripts/build.ts
```diff
@@ -12,6 +12,7 @@ const allTargets = [
   "PyWasmEvaluator",
   "PySvmlEvaluator",
   "PySvmlSinterEvaluator",
+  "PythonStepperEvaluator1",
```
AaravMalani: Maybe rename to `PyStepperEvaluator1` for naming convention sake?

package.json
```diff
@@ -59,7 +59,9 @@
     "typescript-eslint": "^8.56.1"
   },
   "dependencies": {
+    "@sourceacademy/common-stepper": "portal:../plugins/src/common/stepper",
```
AaravMalani: I'm making it a draft till it gets added to the plugins repository

src/conductor/stepper/translate.ts
```diff
@@ -0,0 +1,181 @@
...
+/** Python `repr` for a string: single-quoted with the common escapes. */
+function stringRepr(s: string): string {
+  return `'${s.replace(/\\/g, '\\\\').replace(/'/g, "\\'").replace(/\n/g, '\\n')}'`;
+}
+
+function param(p: FunctionParam): StepNode {
+  return identifier(p.lexeme);
+}
+
+function translateExpr(expr: ExprNS.Expr): StepNode {
```
AaravMalani, line 26: Why have a translation layer? Wouldn't it save a lot of code just by directly operating on the normal AST nodes? (Don't sweat it if it's not feasible, it's pretty minor and I'll still approve without it)

src/conductor/stepper/translate.ts
```diff
@@ -0,0 +1,181 @@
...
+  switch (expr.kind) {
+    case 'BigIntLiteral': {
+      // Python ints are arbitrary-precision; the stepper computes in JS `number`, which is exact for
```
AaravMalani: Err, I don't agree with the cast here. The stepper should compute with BigInts

src/conductor/stepper/translate.ts
```diff
@@ -0,0 +1,181 @@
...
+      const s = stmt as StmtNS.If;
+      return {
+        type: 'IfStatement',
+        test: translateExpr(s.condition),
+        consequent: block(s.body),
+        alternate: s.elseBlock ? block(s.elseBlock) : null,
+      };
+    }
+    case 'Pass':
+      return { type: 'ExpressionStatement', expression: identifier('pass') };
```
AaravMalani: I've not fully gone through the reduction steps, but is this a valid translation?
Even if it does symbolically work, it is a misnomer. Maybe introduce a PassStatement type which does nothing?

src/conductor/stepper/ast.ts
```diff
@@ -0,0 +1,126 @@
+/**
+ * The estree-shaped node model used by the Python substitution stepper.
+ *
+ * The stepper host plugin (`@sourceacademy/web-stepper`) renders steps with a language-agnostic
+ * renderer that dispatches on a node's `type` string and reads estree-style field names
+ * (`left`/`right`/`operator`, `test`/`consequent`/`alternate`, `callee`/`arguments`, ...). To reuse
+ * that renderer unchanged, the Python stepper does not operate on py-slang's class-based AST
+ * directly: it first {@link ../stepper/translate translates} the Python AST into these plain
```
AaravMalani: (Minor) Why the `../stepper/translate` instead of `./translate`?

src/conductor/stepper/reduce.ts
```diff
@@ -0,0 +1,376 @@
...
+ *  - expressions reduce leftmost-innermost (operands first, then the operator/application),
+ *  - `and`/`or`/`if` short-circuit (only the controlling sub-expression is reduced eagerly),
+ *  - applying a `lambda`/single-`return` `def` substitutes arguments for parameters in the body,
+ *  - at the statement level, an assignment or function definition binds a name by substituting its
+ *    value into the following statements (the substitution model's treatment of `const`).
+ *
+ * Each {@link reduceProgram} call performs exactly one contraction and reports the redex/result so
+ * the driver can emit before/after markers. It returns `null` once the program is in normal form.
+ *
+ * Known limitations (acceptable for a teaching stepper): integer arithmetic is computed in JS
```
AaravMalani: These feel like pretty major limitations, hopefully they can be worked on

src/conductor/stepper/reduce.ts
```diff
@@ -0,0 +1,376 @@
...
+  } else if (typeof l === 'number' && typeof r === 'number') {
+    const pyFloat = Boolean(left.pyFloat) || Boolean(right.pyFloat) || op === '/';
+    switch (op) {
+      case '+': result = numberLiteral(l + r, pyFloat); break;
+      case '-': result = numberLiteral(l - r, pyFloat); break;
+      case '*': result = numberLiteral(l * r, pyFloat); break;
+      case '/': result = numberLiteral(l / r, true); break;
+      case '//': result = numberLiteral(Math.floor(l / r), pyFloat); break;
+      case '%': result = numberLiteral(((l % r) + r) % r, pyFloat); break;
+      case '**': result = numberLiteral(l ** r, pyFloat || l ** r !== Math.floor(l ** r)); break;
```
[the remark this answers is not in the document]
AaravMalani: Agreed, please take a look at the existing operator combinations, Python has weird properties (e.g., the `pyMod` function)

It doesn't work exactly the same as JS

AaravMalani (commented):
I've not checked much of the actual reduction, but there a few minor and big changes I'm requesting. Good work anyways!

Shrey5132:
@AaravMalani Thanks for all the feedback
This is a WIP, I only opened the PR to let others see the work I've done so far.
I'm working on all the changes mentioned.

src/conductor/stepper/reduce.ts
Akshay-2007-1, line 1: in / not in / is / is not silently produce "Evaluation stuck" (high)

reduce.ts's contractBinary hits default: return null for these operators — the expression never contracts, and since it's not a value either, the driver reports "Evaluation stuck" with no message. For a teaching tool this is bad UX. Fix: either remove them from operatorPrecedence if unsupported, handle them in contractBinary, or throw a descriptive error.

src/conductor/PyStepperEvaluator.ts
```diff
@@ -0,0 +1,94 @@
...
+   * One-shot run: evaluate the entrypoint, then report completion. We override `startEvaluator`
+   * (rather than only implementing `evaluateChunk`) so that:
+   *  - we emit RUNNING true/false status updates (the host clears the run spinner on RUNNING=false),
+   *  - we emit a terminal STOPPED status so the host's evaluation loop completes and tears down, and
+   *  - we never let the base class send an `undefined` result (which crashes the host saga channel).
+   */
+  override async startEvaluator(entryPoint: string): Promise<void> {
+    const code = await this.conductor.requestFile(entryPoint);
+    if (code === undefined) {
+      this.conductor.sendError(new ConductorError("Cannot load entrypoint file"));
```
Akshay-2007-1, line 46: Missing return after sendError in startEvaluator (low)

if (code === undefined) {
  this.conductor.sendError(new ConductorError("Cannot load entrypoint file"));
  // no return — falls through to sendStatus(STOPPED)
}
STOPPED fires right after ERROR. Likely harmless but could confuse the host state machine. A return after sendError makes intent explicit.

Akshay-2007-1 (changes requested):
The tests for PyCseMachinePlugin.ts (50+) are amazing. I will do the same for CSE Machine Evaluators as well.

DO note that eventually, both of us need to add tests for Conductor integration testing through a mock conductor worker as mentioned at #110.

src/conductor/stepper/reduce.ts
Shrey5132: Thank you for finding and pointing it out. The operators are now completely unsupported and treated as preprocessing/syntax errors.

Akshay-2007-1 (approved):
LGTM

Issue #153, opened 2026-06-12
Specs: CSE machine: loops


martin-henz:
Here is a proposal for handling loops in the Python CSE machine. I didn't check if any or all these details are currently correctly handled, so this issue can serve as a specs that needs to be checked for internal consistency and that the implementation needs to be checked against.

while e : body

leads to the following control

while instruction (with embedded e and body) 
e

The while instruction will pop the value b from the stash. 
Deviation from Python: If b is not boolean, we raise a TypeError.
If b is true, the while instruction will push the following onto control:

while instruction (with embedded e and body) 
e
continue-marker
body

If b is false, the while instruction simply disappears.

for x in range(e1, e2, e3) : body

leads to the following control:

for instruction (with embedded x and body) 
e3 
e2 
e1

The for instruction will pop the values v3, v2, v1 from the stash. If any of them is not an integer, it raises a TypeError.
Deviation from Python: booleans are not considered integers here; if any of them is a boolean, a TypeError is raised. If v3 is 0, a ValueError "range() arg 3 must not be zero" is raised. 
Repeat condition: v1 < v2 for positive v3, v1 > v2 for negative v3.
If the repeat-condition is met, the for instruction will assign x to v1 and push the following onto control:

for instruction (with embedded x and body) 
v3 
v2 
v1 + v3 (as an operator combination)
continue-marker
body

If the repeat condition is not met, the for instruction simply disappears.

break pops instructions from control until (and including) the next for/while instruction. 

continue pops instructions from control until (and including) the next continue-marker.

continue-marker on top of control is simply popped (no-op).

If there is no continue in the body that is outside of a nested loop, the continue marker can be avoided.

Akshay-2007-1:
# Current Status of the CSE Machine in Python via the new Conductor Framework

The status as of now is described here.

## while e : body

For the below code executed in Python - 4.
```python
y = 0
while y < 2 :
    y = y + 1
    
y
```

The CSE Machine looks like this after the assignment of _y_ is done in line 1.

<img width="687" height="448" alt="Image" src="https://github.com/user-attachments/assets/04b15e5a-9ceb-443f-bc66-668029f33b85" />

Does the while instruction pop the boolean value from the stash after the expression is evaluated?
| Before while | After poppin |
|--------|--------|
[...]

PR #155, opened 2026-06-13
Fix CSE machine loop spec compliance (#153)

### LINKED ISSUE #154: Python code not responsing
## Bug: Missing spec compliance from #153 (CSE Machine Loops)

Some of the behaviors specified in #153 are not yet implemented, causing incorrect runtime behavior.

---

### 1. `while` condition does not enforce boolean type

**Spec (#153):** If the condition `b` is not a boolean, raise a `TypeError`.

**Current behavior:** The evaluator uses `isFalsy()` which accepts any type (int, string, None, etc.) as a truthy/falsy value — no error is raised.

**Example:**
```python
y = 0
while y + 1:
    y = y + 1
```
Expected: TypeError (condition is int, not bool)
Actual: Infinite loop — system hangs with no error

---

### 2. `for` v3 = 0 does not raise any ValueError

This is what happens when v3 = 0 currently instead.

https://github.com/user-attachments/assets/8184b91c-4730-40a2-b6d0-ff94b4bd8bb5

The loop statement disappears and no ValueError is raised! 

---

### 3. `for` v1+v3 does not appear individually!

Another thing to note is that currently even when the repeat condition is met, the v1+v3 is not put into the control! Its directly evaluated and the result of that statement is directly put into the control. This can be seen here.

Image

Note that this is the first iteration! Hence the 4 should have been 3 + 1 according to the above specs!

---

Affected file
```
src/engines/cse/interpreter.ts — InstrType.WHILE and InstrType.FOR handlers
```

### Test

Test [here](https://sourceacademy.org/playground) with the feature flag turned on and evaluator selected to Python - 4 with the above code!

Akshay-2007-1:
## What this fixes

Three behaviours specified in #153 were not yet implemented. This PR addresses all three.

---

### 1. `while` — TypeError for non-boolean condition

The spec says the condition must be a boolean; anything else raises a `TypeError`. Previously `isFalsy()` was used directly, which silently accepted integers, strings, `None` etc. This caused code like `while y+1:` to hang indefinitely instead of erroring.

image

---

### 2. `for` — ValueError when step is zero

The spec requires `range() arg 3 must not be zero` to be raised as a `ValueError` when the step argument evaluates to `0`. Previously a zero step fell through the empty-range early-exit condition and produced no output and no error. The early-exit condition was also tightened from `<= / >=` to strict `< / >` to make this work correctly.

image

---

### 3. `for` — v1+v3 pushed as a Binary expression

The spec says the next start value should be pushed onto the control stack as `v1+v3 (as an operator combination)`, not as a pre-computed literal. The increment is now a `Binary` AST node so it appears as an explicit evaluation step in the CSE visualiser.

image

---

## Other changes

Removes the stale `forRangeRole` tagging from `evaluateForIterator` in `utils.ts` — these labels were added for an earlier display approach that has since been replaced and nothing reads them anymore. Refer to the specs at #153 for why this was decided. Utils.ts also has changes related to how the env instructions should be pushed onto the control!

## Files changed

- `src/engines/cse/interpreter.ts`
- `src/engines/cse/utils.ts`

Closes #154

### DIFF
--- a/src/engines/cse/interpreter.ts
+++ b/src/engines/cse/interpreter.ts
@@ -345,6 +345,7 @@ export async function* generateCSEMachineStateStream(
     yield { stash, control, steps };
   }
 }
+
 function isDeclaredEvaluator(kind: string): kind is keyof CmdEvaluators {
   return kind in cmdEvaluators;
 }
@@ -988,15 +989,23 @@ const cmdEvaluators: CmdEvaluators = {
   },
 
   [InstrType.WHILE]: function (
-    _code: string,
+    code: string,
     instr: WhileInstr,
-    _context: Context,
+    context: Context,
     control: Control,
     stash: Stash,
     _isPrelude: boolean,
   ) {
     const condition = stash.pop();
-    if (condition && !isFalsy(condition)) {
+    if (!condition) return;
+    if (condition.type !== "bool") {
+      handleRuntimeError(
+        context,
+        new error.TypeError(code, instr.srcNode as StmtNS.Stmt, context, condition.type, "bool"),
+      );
+      return;
+    }
+    if (!isFalsy(condition)) {
       control.push(instr);
       control.push(instr.test);
       control.push(instrCreator.continueMarkerInstr(instr.srcNode));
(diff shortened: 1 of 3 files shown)


Akshay-2007-1:
@coderabbitai, any comments about the status of the PR?

src/engines/cse/interpreter.ts
```diff
@@ -988,15 +989,23 @@ const cmdEvaluators: CmdEvaluators = {
...
     instr: WhileInstr,
-    _context: Context,
+    context: Context,
     control: Control,
     stash: Stash,
     _isPrelude: boolean,
   ) {
     const condition = stash.pop();
-    if (condition && !isFalsy(condition)) {
+    if (!condition) return;
```
AaravMalani, line 1000: Not at all related to the PR, but should we consider throwing an error if the stash is empty during dev builds @martin-henz @Akshay-2007-1? Again, not an issue for this PR

src/engines/cse/interpreter.ts
```diff
@@ -1029,20 +1038,32 @@ const cmdEvaluators: CmdEvaluators = {
...
       const generateBigIntLiteral = (value: bigint): ExprNS.BigIntLiteral => {
-        const token = new Token(TokenType.BIGINT, value.toString(), 0, 0, 0);
+        const token = new Token(TokenType.BIGINT, value.toString(), 0, 0, -1);
         return new ExprNS.BigIntLiteral(token, token, value.toString());
       };
+      const v1Lit = generateBigIntLiteral(start.value);
+      const v3Lit = generateBigIntLiteral(step.value);
+      const plusToken = new Token(TokenType.PLUS, "+", 0, 0, -1);
+      const nextStartExpr = new ExprNS.Binary(plusToken, plusToken, v1Lit, plusToken, v3Lit);
+      Object.assign(nextStartExpr, { syntheticLabel: `${start.value}+${step.value}` });
```
AaravMalani, line 1063: Is this used for the CSE machine visualisation?

AaravMalani (approved):
LGTM! Just increase the coverage a bit before merging

src/engines/cse/interpreter.ts
Akshay-2007-1: Yes! This is the fix for how @martin-henz stated that the control should look like the following if the boolean condition evals to true
```
for instruction (with embedded x and body)
v3
v2
**v1 + v3 (as an operator combination)**
continue-marker
body
```

Hence, there is this code snippet which is a big clue

```ts
syntheticLabel: `${start.value}+${step.value}
```

AaravMalani: Ah, I was referring to the `syntheticLabel` line. Is that part of the actual CSE machine execution or for tagging it for the visualisation?

src/engines/cse/interpreter.ts
martin-henz: I believe that the stash can't be empty at this point for a syntactically correct program. So the error would be "internal error: report in github".

src/engines/cse/interpreter.ts
Akshay-2007-1: It becomes a binary operation. On the CSE Machine visualization, as shown in the PR content as well, the e1 + e3 operation appears as one control item instead of it being shown after greedy evaluation inside the interpreter/evaluators source file itself.

Then as usual and as expected, the e1 + e3 decomposes into three items
```
+
e1
e3
```

Akshay-2007-1:
> LGTM! Just increase the coverage a bit before merging

I just included more test cases and hence the coverage has gone up

Issue #164, opened 2026-06-23
Stand-alone py-slang implementation for autograding and testing


martin-henz:
For js-slang, we have a stand-alone implementation, which is helpful for testing our systems.

As an example, we are using js-slang to test the textbook SICP JS: We automatically extract the textbook example programs, run them and compare the results with the expected results specified in the textbook XML sources.

Currently, the SICPy testing is done using CPython, which gives us a good ground truth for SICPy; see https://github.com/source-academy/sicp/pull/1254

It would be good if we could also use py-slang for SICPy testing. That way we would get a fairly comprehensive test suite for py-slang for free, namely the SICPy example programs that are already confirmed working with CPython.

For that to work, we would need to deploy py-slang as a headless evaluator. For details on how this is done in js-slang, see https://github.com/source-academy/js-slang/blob/master/README.md#usage

PR #170, opened 2026-06-26
feat: add standalone headless evaluator and Node.js library export

### LINKED ISSUE #169: print() does not append a trailing newline
## Bug

`print()` in py-slang does not append a trailing newline character (`\n`) to its output, unlike CPython where `print()` always ends with a newline by default.

## Found during

This was discovered while building a standalone headless CLI evaluator for py-slang (related to issue #164). When running multiple `print()` calls in sequence, their output is concatenated without any separator:

```python
print(3)
print("hello")
```

**Expected output (CPython behaviour):**
```
3
hello
```

**Actual output (py-slang):**
```
3hello
```

## Root cause

In `src/stdlib/misc.ts`, the `print` implementation passes the formatted string directly to `displayOutput` without appending `\n`:

```typescript
static async print(args: Value[], _source, _command, context): Promise<Value> {
  const output = args.map(arg => toPythonString(arg)).join(" ");
  await displayOutput(context, output);  // no \n appended
  return { type: "none" };
}
```

Compare with the `error()` builtin in the same file, which does append `\n`:
```typescript
const output = "Error: " + args.map(arg => toPythonString(arg)).join(" ") + "\n";
```

## Fix

Append `"\n"` to `output` in the `print` implementation (and respect the `end` keyword argument if that is ever supported).

This is particularly important for correctness when comparing py-slang output against CPython output in automated SICPy testing (issue #164).

Akshay-2007-1:
## Summary

This PR adds two new ways to use py-slang outside the browser:

1. **A standalone CLI** (`dist/repl.cjs`) — run a SICPy file directly from the terminal
2. **A Node.js library export** (`dist/index.cjs`) — import `runCode` from py-slang and call it programmatically

The motivation is issue #164: enabling automated testing of SICPy programs against py-slang itself (instead of CPython), giving us a comprehensive test suite for free from the SICP textbook examples. This is the same pattern js-slang uses — it can be imported as a library by the SICP test runner.

Closes #169.

## Changes

- **`src/runner.ts`** (new): core headless evaluator. `runCode(code, variant)` runs a SICPy program through the CSE machine and returns all stdout as a string. Throws `RunError` on parse/analysis/runtime errors.
- **`src/index.ts`** (new): public library entry point exporting `runCode` and `RunError`.
- **`src/repl.ts`** (refactored): now a thin CLI wrapper around `runCode` — reads a file, calls `runCode`, writes to `process.stdout`.
- **`rollup.lib.mjs`** (new): builds `dist/index.cjs` as a Node.js CJS library bundle.
- **`rollup.repl.mjs`** (new): builds `dist/repl.cjs` as a standalone Node.js CLI.
- **`package.json`**: adds `build:lib` and `build:repl` scripts; moves `commander` to `dependencies`.

## Usage

**CLI:**
```bash
yarn build:repl
yarn repl my_program.py -v 4   # variant 1–4, defaults to 4
```

**Library (e.g. from the SICP test runner):**
```js
const { runCode } = require("/path/to/py-slang/dist/index.cjs");
const output = await runCode("print(1 + 2)", 4);
// output === "3"
```

## Notes

- `runCode` with `maxSnapshots=0` runs the program to completion without collecting CSE snapshots — only stdout is captured. This matches how the chapter 1–2 evaluator path works in `PyCseEvaluator`.
- Discovered during development: `print()` does not append a trailing newline (issue #169).
- The SICP test runner integration is in a separate PR in the `source-academy/sicp` repo, which depends on this PR being merged first.

## Test plan

- [ ] `yarn build:lib` and `yarn build:repl` complete without errors
- [ ] `node dist/repl.cjs <file.py>` runs a SICPy program and prints output
- [ ] `require("dist/index.cjs").runCode(code, variant)` resolves with stdout string
- [ ] `RunError` is thrown on syntax errors and runtime errors
- [ ] Existing evaluator builds (`yarn build`) are unaffected


martin-henz (approved):
looks good!

PR #171, opened 2026-06-27
fix: resolve TypeScript errors blocking yarn build and add prepublishOnly

Akshay-2007-1:
## Summary

- **`PyCseMachinePlugin.ts`**: Fix three type errors in `ControlStackItem` that caused `yarn build` to fail:
  - `body` widened to `unknown` — `WhileInstr.body` is a `StatementSequence` object, not `Array<{kind:string}>` (runtime code already guards with `Array.isArray`)
  - `symbol` typed as `string | TokenType` — unary/binary op instructions carry a `TokenType` enum value, not a plain string; import `TokenType` accordingly
  - `endTok.indexInSource` guarded with `?? 0` to handle the possibly-undefined case (TS18048)
- **`lexer.ts`**: Remove `private` modifier from `PythonLexer.pos` and `PythonLexer.tokens` — TypeScript cannot export a class instance with private members through an anonymous object type (TS4094), which caused rollup-plugin-typescript to error on the `python-grammar.ts` re-export
- **`package.json`**: Add `prepublishOnly` script to run all three build steps (`build --all`, `build:lib`, `build:repl`) automatically before `npm publish`

## Test plan

- [ ] `yarn build --all` completes with no TypeScript errors
- [ ] `yarn build:lib` produces `dist/index.cjs` cleanly
- [ ] `yarn build:repl` produces `dist/repl.cjs` cleanly
- [ ] `tsc --noEmit` reports zero errors
- [ ] `yarn format:ci` passes
- [ ] `yarn lint` reports no new errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "lint": "eslint --concurrency=auto src",
     "format": "prettier --write \"**/*.{ts,tsx,json,js,mjs}\"",
     "format:ci": "prettier --list-different \"**/*.{ts,tsx,json,js,mjs}\"",
-    "wasm": "tsx src/engines/wasm/runFile.ts"
+    "wasm": "tsx src/engines/wasm/runFile.ts",
+    "prepublishOnly": "yarn build --all && yarn build:lib && yarn build:repl"
   },
   "keywords": [
     "Python",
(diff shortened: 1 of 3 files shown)


martin-henz (approved):
Looks good!

PR #174, opened 2026-06-27
fix: remove evaluator build from prepublishOnly

Akshay-2007-1:
## Problem

`prepublishOnly` ran `yarn build --all` which builds all browser evaluator bundles. These bundles include `PyCseMachinePlugin.ts` and `python-grammar.ts`, which emit TypeScript warnings that some environments treat as hard errors, blocking `npm publish`.

## Fix

1. **Remove `yarn build --all` from `prepublishOnly`** — The evaluator bundles are CDN artifacts deployed separately to GitHub Pages, not part of the npm package. `prepublishOnly` now only runs:
   - `yarn build:lib` → `dist/index.cjs` (npm library export)
   - `yarn build:repl` → `dist/repl.cjs` (standalone CLI)

2. **Fix TypeScript warnings** so both the library build and evaluator builds are fully clean:
   - `PyCseMachinePlugin.ts`: fix TS2345 (`rawControlStack` cast) and TS18048 (`endTok.indexInSource` optional chaining)
   - `python-grammar.ts`: prepend `// @ts-nocheck` to suppress TS4094 on the auto-generated nearley parser class; update `python-grammar.ts.sed` so the comment survives future grammar regenerations

Both `yarn build:lib` and `yarn build:repl` now build with zero TypeScript warnings.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -37,4 +37,11 @@ export default defineConfig([
       "@typescript-eslint/no-require-imports": "off",
     },
   },
+  {
+    // Test helpers walk serialized ASTs of arbitrary shape, for which there is no useful static type.
+    files: ["src/**/__tests__/**/*.{ts,tsx}", "src/**/*.test.{ts,tsx}"],
+    rules: {
+      "@typescript-eslint/no-explicit-any": "off",
+    },
+  },
 ]);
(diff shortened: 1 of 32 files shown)


martin-henz (approved):
Looks good!

PR #183, opened 2026-06-30
renaming linked_list to llist

martin-henz:
This PR renames the primitive function linked_list to llist to be consistent with SICPy. 

It also publishes py-slang in npm as 1.0.2.


martin-henz:
this version is already published as 1.0.1 in npm.

Akshay-2007-1 (approved):
LGTM! 
The only caveat I have is that the _linked_list → llist_ removal is an API-breaking change. 

Semver mandates a major version bump (2.0.0). Publishing it as 1.0.2 means any consumer on ^1.0.0 silently picks up a broken API on their next install.

package.json
```diff
@@ -1,9 +1,17 @@
 {
-  "name": "py-slang",
-  "version": "1.0.0",
+  "name": "@sourceacademy/py-slang",
+  "packageManager": "yarn@4.13.0+sha512.5c20ba010c99815433e5c8453112165e673f1c7948d8d2b267f4b5e52097538658388ebc9f9580656d9b75c5cc996f990f611f99304a2197d4c56d21eea370e7",
+  "version": "1.0.2",
+  "publishConfig": {
+    "access": "public",
```
Akshay-2007-1, line 6: Downstream consumers (e.g. source-academy/frontend) with "py-slang": "^1.0.0" in their manifest will silently stop receiving updates with this. 

That should be updated as well if not already there after this merge!

martin-henz: The point is that we don't want downstream packages to silently upgrade, to avoid breaking changes.

PR #185, opened 2026-07-01
feat(stepper): substitute built-in constants and delegate preprocessing to the analyzer

Shrey5132:
## Summary
Two refinements to the conductor Python substitution stepper:

- **Built-in constants** (`math_pi`, …) are substituted into the program before stepping, so they render as their value from the first step (matching js-slang's Source stepper) rather than contracting mid-run — constant identifiers never reach the reducer.
- **Preprocessing** (name resolution + chapter feature-gating) is delegated to the canonical `analyze` pass — the same one the CSE evaluator runs — instead of a bespoke scope walk, so the stepper stays in lockstep with the real evaluator and reports a CPython-style `NameError` for unresolved names. `NameNotFoundError` now carries the offending `varName` for that message.

## Test plan
- `yarn build --evaluator PyStepperEvaluator1` / `PyStepperEvaluator2` build cleanly.
- `getSteps.test.ts` updated and passing.

Companion changes: `source-academy/frontend`, `source-academy/language-directory` (feat/conductor-stepper).


src/conductor/stepper/builtins.ts
```diff
@@ -53,6 +54,22 @@ export function getBuiltinConstant(name: string): StepNode {
...
+ * `program`, as a preprocessing step before stepping. This mirrors js-slang's substitution stepper
+ * (`src/stepper/builtins/index.ts` `prelude`): a constant is not a thing the student reduces, so the
+ * rendered program shows `3.141592653589793` from the very first step rather than displaying
+ * `math_pi` and contracting it later. Substitution is capture-avoiding (it won't touch a `math_pi`
+ * bound as a parameter), so only free references to the constant are replaced.
+ */
+export function substituteBuiltinConstants(program: StepNode): StepNode {
+  let result = program;
+  for (const name of Object.keys(MATH_CONSTANTS)) {
+    result = substitute(result, name, getBuiltinConstant(name));
```
Akshay-2007-1, line 68: This seems well done for the MATH_PI discussion we had in the meeting!

Akshay-2007-1 (approved):
LGTM, approving

PR #192, opened 2026-07-03
Feat/conductor stepper

Shrey5132:
-Implemented support for complex numbers in the python stepper
-Overhauled steps and explanations


src/conductor/stepper/builtins.ts
```diff
@@ -159,6 +164,61 @@ function bankersRound(x: number): number {
...
+    const isImag = s.endsWith("j");
+    if (isImag) {
+      s = s.slice(0, -1);
+      if (s === "" || s === "+" || s === "-") return { real: 0, imag: s === "-" ? -1 : 1 };
+    }
+    if (s in specials) {
+      const v = specials[s];
+      return isImag ? { real: 0, imag: v } : { real: v, imag: 0 };
+    }
+    const n = Number(s);
```
Akshay-2007-1, line 202: parseComplexString's regex split omits the reference parser's limit, silently diverging from the real evaluator for signed real parts.
The reference PyComplexNumber.fromString calls str.split(/(?<!e)(?=[+-])/, 2); the ported version drops the , 2. For complex('-1+2j'), the reference's limited split yields ['', '-1'] → filtered to one part → tries Number('-1+2') → NaN → throws ValueError (rejects the string). The stepper's unlimited split yields ['-1', '+2j'] → two-part branch → succeeds, returning {real: -1, imag: 2}. So the exact same string is accepted by the stepper and rejected by the real evaluator, breaking the PR's own stated invariant ("ported ... so the two evaluators parse the same strings the same way"). Untested — no leading-sign real-part string appears in the new test suite.

src/conductor/stepper/builtins.ts
```diff
@@ -159,6 +164,61 @@ function bankersRound(x: number): number {
...
+  const [realPart, imagPart] = parts;
+  if (!imagPart.endsWith("j")) malformed();
+  const imagStr = imagPart.slice(0, -1);
+  const realOk = realPart in specials || !Number.isNaN(Number(realPart));
+  const imagOk =
+    imagStr in specials || ["+", "-", ""].includes(imagStr) || !Number.isNaN(Number(imagStr));
+  if (!realOk || !imagOk) malformed();
+  const real = realPart in specials ? specials[realPart] : Number(realPart);
+  const imag =
+    imagStr in specials
```
Akshay-2007-1: in on the specials lookup object matches inherited Object.prototype keys, bypassing malformed-string validation.
s in specials, realPart in specials, imagStr in specials all use the bare in operator against a plain object literal, which also matches inherited keys like constructor, toString, valueOf, __proto__. complex('constructorj'): after stripping the trailing j, s === "constructor"; "constructor" in specials is true (inherited), so specials["constructor"] returns the Object constructor function itself, and the code returns {real: 0, imag: <the Object function>} instead of throwing the expected ValueError: complex() arg is a malformed string. The corrupted ComplexValue then propagates (e.g. into complexComponentRepr, which will String()-coerce a function into garbage text) rather than failing cleanly.

src/conductor/stepper/ast.ts
```diff
@@ -42,6 +42,52 @@ export function numberLiteral(n: number, pyFloat: boolean): StepNode {
...
+  if (Number.isNaN(n)) return "nan";
+  if (n === Infinity) return "inf";
+  if (n === -Infinity) return "-inf";
+  if (Math.abs(n) >= 1e16 || (n !== 0 && Math.abs(n) < 1e-4)) {
+    return n.toExponential().replace(/e([+-])(\d)$/, "e$10$2");
+  }
+  return String(n);
+}
+
+/** Python's `repr` of a complex value: `1j` when the real part is zero (no redundant `(...)`/`.0`),
```
Akshay-2007-1, line 77: complexRepr/complexComponentRepr lose the sign of a negative-zero component.
c.real === 0 is true for -0 (JS: -0 === 0), and the fallback String(n) renders -0 as "0" (JS: String(-0) === "0"). Reachable from this same diff: contractUnary's complex branch negates both fields unconditionally ({real: -arg.value.real, imag: -arg.value.imag}), so -(0j) produces {real: -0, imag: -0}. complexRepr then returns "0j", but CPython's actual repr for this value is "-0j" (str(-(0j)) in real Python preserves the sign). Not exercised by any test — the diff's unary-minus tests only use nonzero real parts.

src/conductor/stepper/reduce.ts
```diff
@@ -617,7 +774,10 @@ function stepHead(head: StepNode, rest: StepNode[]): HeadOutcome {
         kind: "step",
         newBody: rest.map(stmt => substitute(stmt, name, boundValue)),
         preRedex: head,
-        explanation: `${name} declared and substituted into the rest of the program`,
+        postRedex: head,
+        postNewBody: [head, ...rest],
+        explanation: `Declared and substituted ${name} into the rest of the block`,
+        beforeExplanation: `Declaring and substituting ${name} into the rest of the block`,
       };
     }
```
Akshay-2007-1, line 786: FunctionDeclaration and plain NameBinding now produce byte-identical explanations, losing the distinction between binding a function and binding a value.
The old text differed ("Function X declared, parameter(s) Y required" vs. "X declared and substituted into the rest of the program"); both branches now emit the exact same Declared and substituted ${name} into the rest of the block (and matching beforeExplanation). A student stepping through def square(n): ... and through x = 5 sees identical wording for both, even though one is declaring a callable with parameters and the other a plain value — a real loss of pedagogical signal in a stepper whose whole purpose is showing what kind of step just happened. (The PR's own updated tests accept this merged wording, so this may be intentional — worth a maintainer's explicit sign-off rather than an accidental regression.)

Akshay-2007-1 (changes requested):
A few changes caught by Claude as follows :

src/conductor/stepper/reduce.ts
Shrey5132: This is an intentional design choice. I wanted to make it clear that functions are treated the same as variables in the functional programming paradigm.

Akshay-2007-1 (approved):

PR #194, opened 2026-07-04
fix(stepper): report preprocessing errors identically to the default evaluator

Shrey5132:
## Summary
- Running a program with a preprocessing/syntax error (e.g. an undefined name) reported a different message depending on which evaluator caught it: the Stepper tab showed a simplified `NameError: name 'x' is not defined`, while the default CSE evaluator showed the analyzer's full formatted diagnostic (line number, source context, `^^^` underline, and a "Perhaps you meant..." suggestion).
- `preprocessPython` already delegated name resolution to the exact same `analyze()` pass the CSE evaluator runs — that part was already shared — but then discarded the analyzer's own message for the specific case of an unresolved name, rebuilding a simplified string instead. Every other analyzer error (chapter feature-gates, forbidden reassignment, etc.) already used the real message unmodified; only this one case diverged.
- Fix: return `errors[0].message` unconditionally. A preprocessing error now reads identically regardless of which evaluator caught it — verified the exact reported repro (`asd`) produces byte-for-byte the same output in both.
- Removed `NameNotFoundError.varName`: it existed solely to build the now-deleted simplified string and had no other callers.

## Test plan
- Updated the 13 existing stepper preprocessing tests in `getSteps.test.ts` that asserted the old simplified string, to the analyzer's real formatted message (independently re-derived each expected string via a throwaway probe before locking it into the test).
- Full suite: 2448/2448 passing.
- `tsc` clean.
- Manually verified `asd` produces the exact reported CSE-evaluator output through the stepper's preprocessing path.

### DIFF
--- a/src/conductor/stepper/__tests__/getSteps.test.ts
+++ b/src/conductor/stepper/__tests__/getSteps.test.ts
@@ -255,7 +255,12 @@ describe("Python stepper — built-in functions and constants", () => {
 
 describe("Python stepper — undefined variables are a preprocessing error", () => {
   test("an undefined name is reported (and would block the stepper)", () => {
-    expect(preprocess("undefined_name")).toBe("NameError: name 'undefined_name' is not defined");
+    // Same message the default (CSE) evaluator reports for the same program: the analyzer's own
+    // formatted `NameNotFoundError`, not a stepper-specific simplification.
+    expect(preprocess("undefined_name")).toBe(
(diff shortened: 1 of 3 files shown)


Akshay-2007-1 (approved):
LGTM!

One non-blocking note for the PR description: it bundles an unrelated dead-code deletion (BreakContinueError) that isn't mentioned in the summary, worth calling it out.

PR #196, opened 2026-07-04
feat(cse-machine): serialize globalNames onto call frames + fix premature return-frame transition

### LINKED ISSUE #4073: CSE Machine (Python): frame transition on return happens a step early, at the return statement instead of the ENVIRONMENT instruction
## Description

When stepping out of a function call in the Python CSE machine visualization, the active-frame transition back to the caller happens at the `return` statement's own step, instead of at the subsequent `ENVIRONMENT` instruction's step — which is where `EnvironmentAnimation` is supposed to animate that transition (`CseMachineAnimation.tsx`, `case InstrType.ENVIRONMENT`).

## Root cause (py-slang)

py-slang's `[InstrType.RESET]` handler (`src/engines/cse/interpreter.ts`) called `popEnvironment(context)` directly. In js-slang's real CSE machine, `RESET` never touches the environment — only `ENVIRONMENT` does, via a `while` loop that pops back to the target env. Since `ENVIRONMENT` is always pushed immediately below `RESET` for every function call (`APPLICATION` handler pushes `envInstr` then `resetInstr`), popping the environment in `RESET` made the transition happen one step early; by the time `ENVIRONMENT` executed, its `while` loop condition was already satisfied (no-op), so there was nothing left to animate there.

## Impact

This affects every function return in the Python CSE machine — the frame highlight snaps back to the caller instantly at the return statement rather than animating at the intended step, which matters for the same pedagogical reason as #4041/#4072: tracing frame transitions (including through `global`/`nonlocal` scopes) is how students follow what the CSE machine is actually doing.

## Fix

The entire fix lives in py-slang, not this repo — no frontend code changes are needed. Fixed in source-academy/py-slang#200: removes the redundant `popEnvironment(context)` call from the `RESET` handler, leaving environment restoration entirely to the paired `ENVIRONMENT` instruction (matching js-slang's design). Verified via a new regression test asserting the callee frame is still active immediately after the return step, and only stops being active once the `ENVIRONMENT` instruction's step has run. Full py-slang test suite (2449 tests) passes unaffected.

(Tracked here rather than only in py-slang since it's part of the same CSE machine visualization work as #4041/#4042/#4072.)

Akshay-2007-1:
## Summary
- Expose `Closure.globalVariables` (already computed at function-def time and already used by the interpreter to route `global`-declared name lookups straight to the global frame) via the new optional `CseSerializedEnvFrame.globalNames` field, so the visualizer can annotate call frames the moment they're created.
- Fixes source-academy/frontend#4073: `RESET`'s handler called `popEnvironment(context)` directly, redundant with `ENVIRONMENT` (always pushed immediately below `RESET` by `APPLICATION`) which already fully restores the environment via its own `while` loop. This made the CSE machine visualizer's frame transition on return happen one step early, at the return statement itself, instead of at the `ENVIRONMENT` instruction where `EnvironmentAnimation` animates it. Removed the redundant pop, matching js-slang's design (RESET never touches the environment there).

(Originally two separate PRs — #200 folded in here via cherry-pick so only one review is needed. #200 is now closed.)

Depends on the `globalNames` field added to `@sourceacademy/common-cse-machine` in the companion source-academy/plugins PR; the runtime behavior here is safe regardless (extra JSON property), but the test file casts at the read site until py-slang's dependency on that package is bumped past the version declaring the field.

Part of source-academy/frontend#4041, source-academy/frontend#4073.

## Test plan
- [x] `yarn jest` — 2426/2426 passing
  - 3 new tests for `globalNames` in `serializeEnvChain`
  - 1 new regression test: callee frame stays active through the return step, only stops being active once `ENVIRONMENT`'s step has run
- [x] `yarn lint` — no new errors (pre-existing unrelated warnings only)
- [x] `npx prettier --check` on touched files — clean


src/conductor/plugins/PyCseMachinePlugin.ts
```diff
@@ -390,6 +390,7 @@ function serializeEnvChain(
         })),
       isActive: env.id === activeEnv.id,
       isOnCallStack: callStackIds.has(env.id),
+      globalNames: env.closure?.globalVariables.size ? [...env.closure.globalVariables] : undefined,
```
Akshay-2007-1, line 393:
```suggestion
     (globalNames: env.closure?.globalVariables !== undefined) && globalNames: env.closure?.globalVariables.size ? [...env.closure.globalVariables] : undefined,
```

Shrey5132 (approved):

PR #197, opened 2026-07-04
feat(stepper): support breakpoint() as a debugger breakpoint

### LINKED ISSUE #188: Stepper: breakpoint() function enables big steps
In Python, we should support the builtin breakpoint. So
breakpoint()
in Python should play the role of
debugger;
in JavaScript.

Image

Shrey5132:
## Summary

Implements `breakpoint()` support in the Python substitution stepper (Closes #188). It is the stepper's analogue of JavaScript's `debugger;`: for code evaluation it does nothing and behaves exactly like `pass`, but it introduces a breakpoint the stepper's navigation (the double-arrow / "next breakpoint" control) can jump to.

Per the issue's spec, the double-arrow lands on the step whose explanation is **"Evaluating breakpoint statement"**, never the following **"Evaluated breakpoint statement"** step — even though (like `pass`) the statement is still shown, highlighted green, on that after step before it disappears.

## How it works

The host stepper is language-agnostic and recognises a breakpoint by a marker whose `redexNodeType === "DebuggerStatement"`, so no host/plugin change was needed:

- **`translate.ts`** — a no-arg `breakpoint()` statement is translated to the shared `DebuggerStatement` node. (Used as an expression, e.g. `x = breakpoint()`, it stays an ordinary call to the built-in.)
- **`reduce.ts`** — the `DebuggerStatement` contraction drops the statement like `pass`, keeping it highlighted green on its after step (`postRedex`/`postNewBody`).
- **`getSteps.ts`** — `redexNodeType` is now emitted on the **before** marker only. It names the redex *about to be* contracted and drives breakpoint navigation, so it belongs on the before step; an after marker still carries `redexId`/`redexType` (so the statement is highlighted green) but is no longer a navigation target. This is what makes the double-arrow stop only on "Evaluating breakpoint statement".
- **`builtins.ts`** — `breakpoint` is registered as a no-op returning `None`, so the name resolves during preprocessing (it joins `getAvailableBuiltinNames`) and degrades gracefully in expression position.
- **`syntaxProfile.ts`** — renders `DebuggerStatement` as `breakpoint()`.

## Testing

- Added a `breakpoint()` test suite (preprocessing acceptance, no-op semantics in top-level/binding/function-body positions, the exact step/explanation sequence, before-only breakpoint-nav targeting across multiple breakpoints, the green flash on the after step, and expression-position graceful degradation).
- Full py-slang suite: **2455 passed** / 21 suites. `tsc`, ESLint, and Prettier all clean.


src/conductor/stepper/translate.ts
Akshay-2007-1, line 1: Shadowing breakpoint is silently ignored (CONFIRMED, verified by running against the branch)

For the following code : 
```python
bnt = 5
bnt()
```

The expected behaviour is that there should be an error and we be redirected to the home tab like any other invalid code. This is what happens, say, for the CSE Machine as well for the same code in Python 3 and 4.

For Stepper specifically, the code actually evaluates and gets here : 

Image

---

Moreover, every other builtin (print, error, etc.) is resolved normally through the environment at evaluation time and correctly respects shadowing, breakpoint is the only name special-cased ahead of scope resolution, which is also why an independent altitude-angle check flagged this as implemented one layer too shallow (pattern-matched in the translator instead of resolved via the analyzer/environment that already handles this for every other name).

src/conductor/stepper/getSteps.ts
Akshay-2007-1, line 1: redexNodeType is now dropped from every after-marker in the stepper, not just DebuggerStatement's (PLAUSIBLE).
The old line unconditionally set out.redexNodeType = marker.redex.type for any marker with a redex; the PR gates it to before-markers only, to stop the double-arrow from also landing on "Evaluated breakpoint statement." That gate is global, so it also strips redexNodeType from after-markers for PassStatement, IfStatement, VariableDeclaration, etc. The shared type's doc comment only says the field is "used e.g. for breakpoint navigation" (non-exhaustive), so if the external host (a separate repo, not visible here) reads redexNodeType off after-markers for anything besides breakpoint-nav (e.g., type-based highlighting/analytics on the post-reduction step), it would silently lose that data for every statement type, not just the new one. Nothing in this repo's tests exercises that case either way — worth a quick check against the host/frontend repo before merging.

src/conductor/stepper/reduce.ts
Akshay-2007-1, line 1: The new DebuggerStatement case is a byte-for-byte structural duplicate of the PassStatement case directly above it (same newBody/preRedex/postRedex/postNewBody shape), differing only in the two explanation strings. Minor maintenance cost — a shared "no-op, green-flash-before-discard" helper keyed by explanation strings would keep this from drifting if a third such statement is added later. Not blocking.

Akshay-2007-1 (changes requested):
Image

```python
def breakpoint():
    return 42

breakpoint()
```

As you can see, it takes even a function called breakpoint() as an acutal breakpoint statement though that may not have been the case. I think there should be restriction that a function cant have the name "breakpoint" as its a keyword. @martin-henz, what do you think? Is this too strict to check and we can leave it be?

martin-henz:
> Image
> ```python
> def breakpoint():
>     return 42
> 
> breakpoint()
> ```
> 
> As you can see, it takes even a function called breakpoint() as an acutal breakpoint statement though that may not have been the case. I think there should be restriction that a function cant have the name "breakpoint" as its a keyword. @martin-henz, what do you think? Is this too strict to check and we can leave it be?

I think we have a misunderstanding here @Akshay-2007-1 . Python has a builtin function `breakpoint`. There is no keyword **breakpoint** in Python. The name `breakpoint` is predeclared (just like `print`) and get treated specially by the implementation (just like `print`). So this should work:
```
bp = breakpoint
bp()    # should create a breakpoint in the stepper/CSE machine

**Comment by Akshay-2007-1** [comment]
> I think we have a misunderstanding here @Akshay-2007-1 . Python has a builtin function `breakpoint`. There is no keyword **breakpoint** in Python. The name `breakpoint` is predeclared (just like `print`) and get treated specially by the implementation (just like `print`). So this should work:
> 
> ```
> bp = breakpoint
> bp()    # should create a breakpoint in the stepper/CSE machine
> ```

"keyword" was the wrong word, and I shouldn't have suggested disallowing the name.

But I think your own example actually shows the underlying issue more clearly than mine did:

```python
bp = breakpoint
bp()    # should create a breakpoint in the stepper/CSE machine
```

For this to work, the implementation needs to recognize a call to the **value** bound to the builtin `breakpoint` function, wherever that value ends up (`bp`, or any other alias). Right now `translate.ts` doesn't do that: it pattern-matches the **literal spelling of the identifier** (`callee.name.lexeme === "breakpoint"`) before any name/scope resolution happens. That means today, as implemented:

- `bp = breakpoint; bp()` -> does **not** trigger a breakpoint (misses your case, since the callee is spelled `bp`, not `breakpoint`)

image
Pressing the double chevron to reach breakpoints DOES NOT happen.

---

- `def breakpoint(): return 42` then `breakpoint()` — **does** trigger a breakpoint, even though the name has been rebound to a completely different function and the builtin is no longer reachable under that name

image
Pressing the double chevron to reach breakpoints DOES happen while it aint supposed to. So my earlier suggestion was throw an error on an attempt to define a function with the name "breakpoint"

---

Both are the same root cause: the check happens on the identifier's spelling instead of on what it actually resolves to. `print`/`error` don't have this problem because they're resolved normally through the environment at evaluation time, so shadowing (and presumably aliasing) is handled correctly for them already.

@martin-henz, @Shrey5132 could change the logic from plain spelling search into normal resolution like the case of `print` or other built-ins.

src/conductor/stepper/translate.ts
Shrey5132: This is intended behaviour, since the python evaluator does not know at the time of preprocessing that bnt is a function.

src/conductor/stepper/reduce.ts
Shrey5132: I don't think a new no-op statement will be added in the future, and also there is no real reason to keep breakpoint() coupled with pass. We may want to tweak the functionality of breakpoint() separate from pass in the future.

src/conductor/stepper/getSteps.ts
Shrey5132: Checked this against the actual host repo (plugins) — not an actual issue.

`redexNodeType` has exactly 3 occurrences in the whole plugins codebase: the type definition itself (`common/stepper/src/index.ts`), and two usages in `SubstVisualizer.tsx` — `stepNextBreakpoint`/`stepPreviousBreakpoint`, both checking `marker.redexNodeType === "DebuggerStatement"` exclusively. Nothing reads it for highlighting, analytics, or any type-based rendering.

A few more data points:
- Git history: the field was introduced into the plugins repo in the *same commit* that added `stepNextBreakpoint`/`stepPreviousBreakpoint` (`96c1eaf`/#20, "add common-stepper and runner-stepper plugin packages"). Searching all commits/branches in that repo for `redexNodeType` turns up nothing else, ever — it's never had a second consumer.
- `frontend` has zero references to `redexNodeType`/`SerializedMarker`/`SerializedStepperStep` — it only embeds the plugin bundle, it doesn't parse the protocol directly.
- `js-slang` doesn't depend on `common-stepper` at all; its stepper is a separate, pre-conductor legacy implementation, unaffected either way.

So the doc comment's "used *e.g.* for breakpoint navigation" is non-exhaustive in wording, but in practice there's no other case — the field was purpose-built for this one feature. The `main`-branch unconditional emission (before this PR) looks like an unexamined default from the original stepper scaffold rather than a deliberate contract anything relies on. Agreed it's worth tightening that doc comment to state the before-marker/breakpoint-nav-only contract explicitly so this doesn't need re-litigating later, but that's a separate small change in the plugins repo, not blocking here.

Akshay-2007-1 (approved):
LGTM! I tested and it works!

PR #201, opened 2026-07-05
fix(stepper): substitute a binding into the rest of the block on its own step

Shrey5132:
## Summary

Consistency fix for the substitution stepper: a name binding's substitution into the rest of the block was visible one step *after* the step that claims to have already performed it.

For:
```python
def f(x):
    y = x + 1
    return y
f(1)
```
the step explained **"Declared and substituted y into the rest of the block"** (past tense — implying it already happened) still showed `return y` — the substitution of `y` only became visible on the *following*, unrelated step. This didn't match how function calls already behave: `f(1)`'s argument substitution is visible immediately on its own **"Substituted 1 into x of f"** step (`contractCall`), not the next one.

## Fix

`stepHead`'s `VariableDeclaration`/`FunctionDeclaration` cases (`reduce.ts`) now compute the substituted rest once and reuse it for both:
- `newBody` — the tree the next contraction starts from (unchanged behaviour), and
- `postNewBody` — the tree *this* step's own after (green) display uses (previously the unsubstituted `rest`).

The declaration itself still lingers, highlighted green, for exactly one more step before disappearing (the same "green flash before discard" `pass` gets) — only the timing of the substitution's effect on the rest of the program moves earlier, to the step whose explanation already claims it happened.

## Testing

Added a test suite ("a binding's substitution is visible on its own step, not the next") covering: a top-level variable binding, a top-level function declaration, and the local-binding-inside-a-function-body case from the report above. Confirmed each new test fails without the `reduce.ts` change (temporarily reverted it to check) before finalizing.

Full py-slang suite: **2451 passed** / 21 suites. `tsc`, ESLint, and Prettier all clean.

### DIFF
--- a/src/conductor/stepper/__tests__/getSteps.test.ts
+++ b/src/conductor/stepper/__tests__/getSteps.test.ts
@@ -673,6 +673,78 @@ describe("Python stepper — function values render as mu-terms (not inline bodi
   });
 });
 
+describe("Python stepper — a binding's substitution is visible on its own step, not the next", () => {
+  // A name binding (`VariableDeclaration`/`FunctionDeclaration`) substitutes its value into the rest
+  // of the block on the *same* contraction that declares it — matching how a call expression's
+  // argument substitution is already visible on its own "Substituted ..." step (`contractCall`), rather
+  // than only becoming visible one (unrelated) step later. The declaration itself still lingers one
+  // more step, highlighted green, before actually disappearing — the same "green flash before discard"
+  // `pass` gets — but that lingering must not delay the substitution's effect on the rest of the tree.
+
+  test("a variable's bound value already appears in the rest on its own 'Declared and substituted' step", () => {
+    const s = steps("x = 5\nx + 1");
+    const i = s.findIndex(
+      step =>
+        step.markers?.[0]?.explanation === "Declared and substituted x into the rest of the block",
+    );
+    expect(i).toBeGreaterThan(-1);
+
+    // The declaration is still present (green flash, like `pass`)...
+    expect(findNode(s[i].ast, n => n.type === "VariableDeclaration")).toBeDefined();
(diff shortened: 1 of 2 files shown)


Shrey5132:
## `coverage/coveralls` failure analysis

The `coverage/coveralls` check reports "Coverage decreased (-0.004%) to 78.683%", but this does not appear to be a real regression introduced by this PR:

1. **Coveralls' own diff view** reports the 2 new/added lines in this PR's one changed source file are **100% covered**.
2. **Mathematically**, adding only fully-covered lines to a codebase can only raise or hold flat the aggregate coverage percentage — it cannot cause a decrease.
3. I reproduced the exact LCOV data (what Coveralls ingests) locally, running the full test suite against both `main` and this branch. Every file's covered/missed line counts came out **byte-identical** between the two runs — including `reduce.ts`, the only file with substantive changes in this PR.
4. I re-ran the CI workflow once (same commit) — it reproduced the identical `-0.004%` deterministically on the GitHub Actions runner, but I could not reproduce any difference at all locally, which points to something specific to the CI environment rather than to this diff.
5. Sibling PRs based on the same `main` commit (#197, and the earlier #194/#192) all show coverage **increases**, consistent with normal run-to-run fluctuation rather than a regression pattern tied to this change.

`build` (tests, lint, format) all pass cleanly. Only the Coveralls threshold gate — which appears to fail on any decrease with no tolerance margin — is red, by a statistically negligible amount (2 lines out of 7361 relevant lines repo-wide).

There's no in-repo `.coveralls.yml` or threshold config to adjust; the pass/fail policy is controlled on Coveralls' dashboard, outside this repo. Flagging this for a maintainer with Coveralls access to review/override, since I don't believe there's a code change on this branch that would legitimately resolve it.

martin-henz (approved):
looks good!

PR #203, opened 2026-07-05
feat(stepper): report cumulative print output per step

Shrey5132:
## Summary

Makes the Python stepper report the program's **cumulative output** (everything `print` has written) on each step, so the stepper host can show a running output panel that grows as the slider advances. This is the **runner** half; the **host** half (rendering the panel) is a companion PR in [source-academy/plugins](https://github.com/source-academy/plugins).

## Behaviour
- A `print(...)`'s text first appears on its **"Ran print"** step and persists on every step after it (cumulative — later steps repeat all earlier output plus their own).
- Formatting mirrors CPython defaults: `print(a, b)` → `a b\n`, `print()` → a blank line, args rendered with `str()`. (`print`'s keyword args — `sep`/`end` — aren't modelled; the AST carries only positional args.)
- `print` still evaluates to `None`.
- A program that never prints carries no `output` field on any step.

## Changes
- **`builtins.ts`** — `formatPrintOutput(args)` helper.
- **`reduce.ts`** — `ReduceResult.output`; `contractCall` records a print call's text; threaded through `stepHead`/`reduceProgram`/`reduceBlock`.
- **`getSteps.ts`** — `drive()` accumulates output across contractions and stamps every step; serialized out (emitted only when non-empty).

## A note on the type
`output` belongs on the shared `SerializedStepperStep` protocol type, which the plugins PR adds to `@sourceacademy/common-stepper` (published as `0.0.2`). Until this package bumps that dependency, the field is emitted through a small local type-widening (marked with a `TODO`); it crosses the runner→host channel as plain JSON regardless of the static type, and is forward/backward compatible (host ignores it if absent; an old host ignores it if present). So this PR and the plugins PR are independently mergeable in any order.

## Testing
Added a `cumulative print output` test suite (output on "Ran print" not "Running print", cumulative across prints and function bodies, CPython formatting, persistence to "Evaluation complete", and no field when nothing prints) — each confirmed to fail without the change. Full suite: **2454 pass** / 21 suites. `tsc`, ESLint, Prettier clean.

> Note: touches `reduce.ts`, which open PRs #197 and #201 also touch (different areas) — whichever merges later needs a trivial rebase.


Shrey5132:
Note that CI coverage tests can be ignored for this PR since I have added the test cases, but there are some issues with coveralls bot.

src/conductor/stepper/builtins.ts
```diff
@@ -134,6 +134,32 @@ function pyStr(node: StepNode, repr: boolean): string {
...
+ * so those defaults are fixed. The reducer records this as the output a `print(...)` contraction adds
+ * to the stepper's cumulative output panel; `print` itself still evaluates to `None` (its Python return
+ * value — see the `print` built-in below).
+ */
+export function formatPrintOutput(args: StepNode[]): string {
+  return args.map(a => pyStr(a, false)).join(" ") + "\n";
+}
+
+/** Python truthiness for the value subset the stepper handles (mirrors `reduce.ts`'s `isTruthy`). */
+function truthy(node: StepNode): boolean {
```
Akshay-2007-1: dead, duplicate truthy() function. The diff adds a truthy(node) helper whose own doc comment says it "mirrors reduce.ts's isTruthy" — and it is in fact a line-for-line duplicate of isTruthy (reduce.ts:87-98). I confirmed via grep that truthy is never called anywhere in the file or the diff. It's unreferenced logic that will silently drift from isTruthy if Python truthiness semantics change later. tsconfig.json's noUnusedLocals is commented out and ESLint's no-unused-vars is only a warn, so this won't fail CI, but it's still dead code that should either be deleted or wired up (or isTruthy exported from reduce.ts and imported here instead of copied).

src/conductor/stepper/getSteps.ts
```diff
@@ -64,8 +81,12 @@ function drive(prog: StepNode, contractionLimit: number): Step[] {
...
       // mirroring Source (which ends a failed run with "Evaluation stuck" rather than "complete").
       const message = error instanceof Error ? error.message : String(error);
-      steps.push({ ast: current, markers: [{ redexType: "beforeMarker", explanation: message }] });
-      steps.push({ ast: current, markers: [{ explanation: "Evaluation stuck" }] });
+      steps.push({
+        ast: current,
+        markers: [{ redexType: "beforeMarker", explanation: message }],
+        output,
+      });
+      steps.push({ ast: current, markers: [{ explanation: "Evaluation stuck" }], output });
```
Akshay-2007-1: repeated output stamping across 7 steps.push(...) call sites. Every push in drive() (the initial step, two error-path pushes, the terminal-null push, before/after-contraction pushes, and the max-steps-exceeded push) manually re-adds output, to the pushed object. A small pushStep(ast, markers) closure capturing output by reference would remove this copy-paste and the risk of a future new push site forgetting to stamp it.

Akshay-2007-1 (changes requested):
Two minor style picks, nothing blocking correctness per se.

I am only requesting changes and not approving as the `trythy()` bug was also flagged by the CI as a warning.

src/conductor/stepper/getSteps.ts
martin-henz: noted. no concern.

src/conductor/stepper/builtins.ts
Shrey5132: Good catch on the duplication risk, though `truthy` wasn't actually dead at the time — it backed the `bool()` builtin's conversion (`bool(x)` → `truthy(x)`), so that call site must have slipped past the grep somehow. Either way, the fix you suggested is the right one: hoisted a single `isTruthy` predicate into `ast.ts` (the module both `reduce.ts` and `builtins.ts` already depend on) instead of having `builtins.ts` import it from `reduce.ts`, which would've created an import cycle (`reduce.ts` already imports `builtins.ts`). Both call sites now share the one implementation.

Separately: the `main` sync since your review removed the `bool`/`int`/`float` builtins entirely, so `truthy`'s only call site is gone regardless now — but the dedup holds either way.

Fixed in 73ca1db.

src/conductor/stepper/getSteps.ts
Shrey5132: Done — replaced all 7 call sites with a `pushStep(ast, markers)` closure that captures `output` by reference, so it's stamped in exactly one place and a future push site can't forget it.

Fixed in 73ca1db.

Akshay-2007-1 (approved):

Issue #206, opened 2026-07-06
Bring the SVML engine in line with the revised operator semantics


martin-henz:
Follow-up to #198 (sibling issues for the other engines: WASM #205, stepper #212). The `operators` branch revises the operator typing spec (`docs/specs/python_typing_{front,middle_1,middle_2,middle_34,back,tail_2,tail_34}.tex`) and implements it in the **CSE engine**, with a spec-driven conformance sweep (`src/tests/operator-conformance.test.ts`). The **SVML engine** (`src/engines/svml/`) has its own operator implementation and was not touched — it needs the same semantics.

**Update**: §1 and §2 are no longer identical (the original version of this issue described them as one unchanged block) — §2 gained its own, more permissive `==`/`!=` rule after this issue was first filed. See the corrected breakdown below.

## What changed in the CSE engine (the reference behavior)

At **Python §1** (unchanged):
- `==` / `!=` only numeric × numeric and string × string (no bool operands, no cross-type, no pairs/None).
- `>,>=,<,<=` only int/float × int/float and string × string.
- `is` / `is not` rejected at validation (`NoIsOperatorValidator` — note: this validator was broken and never fired; fixed on the branch to match `ExprNS.Compare` and `TokenType.ISNOT`).

At **Python §2** (revised — distinct from §1):
- `==` / `!=` are now structural over **any × any**, *except* that bool and function are excluded entirely (`bool,function x any -> error`, `any x bool,function -> error`) — a §2 comparison never has to answer whether `True == 1` holds, or what function equality means before `is` is introduced at §3/§4. Every other combination is `bool`, including cross-type comparisons (`1 == 'abc'` → `False`, not an error) and pairs/None (`pair(1, 2) == pair(1, 2)` structural; `pair(1, 2) == None` → `False`).
- `>,>=,<,<=` unchanged from §1: only int/float × int/float and string × string (bool still excluded from ordering at §2).
- `is` / `is not` still rejected at validation, same as §1 (the operator only exists from §3 on).

At **Python §3/§4**:
- `==` / `!=` take **any × any**: structural (element-wise, recursive) equality on lists, and booleans compare as the ints they are, as in CPython (`True == 1`, `[True] == [1]`).
- `is` / `is not` **error** whenever either operand is a number, string or bool (identity of immutables is unobservable — the interning trap); on **list / function / None** they compare by reference (`xs is None` valid; `None is None` true).
[...]

PR #207, opened 2026-07-06
CSE machine: Spec-driven operator semantics for Python §1–§4: ==, is, ordering (#198)

### LINKED ISSUE #198: Spec-driven operator-restriction tests + factor the typing tables with \input (after #193)
> **Ordering:** do this *after* #193 lands. It edits `docs/specs/python_typing*.tex` and adds operator tests that overlap `getSteps.test.ts` / `stdlib.test.ts`, so it would conflict with the rename+removal PR.

Two related items came out of reviewing the operator type spec (`docs/specs/python_typing.tex` = Python §1/§2, `python_typing_3.tex` = §3/§4): the two spec tables have drifted from each other by hand, and nothing systematically checks that the implementation matches the spec's operator **restrictions**.

## Part A — Factor the typing tables with `\input`

The two files are independent full copies of the operator table. Diffing them, the *intended* difference is small, but they also differ **accidentally**:

**Intended semantic delta**
- `==` / `!=`: §1/2 = `numeric×numeric`, `string×string`; §3/4 = `any × any → bool` **plus** `list × list → error`.
- `is`: §3/4 only (`is  list list → bool`).
- Rationale: §1/2 are strictly referentially transparent (sharing of data structures is unobservable), so structural equality is restricted. §3/4 model sharing, so `==` becomes total — except `list × list`, kept an error so sharing is not exposed.

**Accidental drift (should be fixed)**
- Common rows are formatted differently (§1/2 groups `+,-,*,%` and `>,>=,<,<=`; §3/4 splits them per operator), so they are not byte-identical.
- **§3/4 is missing the `% float × int → float` row** that §1/2 has — a live inconsistency found by hand-diffing.

**Proposal.** Split the table so the common rows live once and only the differing rows are per-chapter. `\input` works inside `tabular` (it is input-stream insertion, so `\begin{tabular}` / rows / `\end{tabular}` may span files; just no blank lines inside a fragment, and use `\input` not `\include`):

```
python_typing.tex    ->  \input{python_typing_front} \input{python_typing_middle_12} \input{python_typing_back}
python_typing_3.tex  ->  \input{python_typing_front} \input{python_typing_middle_34} \input{python_typing_back}
```
- `front` = `\begin{tabular}` + header + common top rows
- `middle_12` / `middle_34` = the rows that differ (`==`/`!=`, and `is` for §3/4)
- `back` = common bottom rows + `\end{tabular}`

The real work is the one-time reconciliation that makes a single shared body possible: normalize the common rows to one formatting, fix the `% float×int` omission, and group the chapter-specific rows contiguously (or `\input` them at their positions with an empty stub for §1/2). Afterwards the diff between `middle_12` and `middle_34` **is** the documented §1/2→§3/4 change, and the common rows cannot drift.

## Part B — Spec-driven operator-restriction tests

The operator implementation is imperative `switch`/`case` with no shared type table, and today's operator tests (`stdlib.test.ts`: `add, sub, mul` / `div and mod` / …, and the stepper's `getSteps.test.ts`) are hand-enumerated **positive** cases. Nothing systematically checks that every **forbidden** combination errors — that negative space *is* the restrictions, and it is where bugs hide.

**Proposal.** A table-driven conformance test: encode the spec table once as `(op, leftType, rightType, variant) -> resultType | error`, then generate the full cross-product of `operator × type × type`:
- allowed -> assert the result **type** (value correctness is already covered elsewhere);
- everything else -> assert `UnsupportedOperandTypeError`.

A prototype (arithmetic + comparisons, ~30 lines of table -> 396 generated cases) already works and immediately surfaced the **chapter dependence**: run at variant 3, the §1/2 table flags every `==`/`!=` mixed-type case, because the engine (correctly, per `python_typing_3.tex`) makes `==` total at §3/4. So the test **must be parameterized by variant** and use the chapter-appropriate table. Useful side effect: a variant-parameterized test would automatically answer *"does the engine actually enforce the §1/2 `==` restriction at variant 1?"* — a conformance question we have not verified.

Cases needing care: `**` (result type is value-dependent: `int**int>=0 -> int`, `int**int<0 -> float`), `and`/`or` (restriction is a bool **left** operand; result is value-dependent), `not`, unary `-`. Stretch goal: run the same table against the CSE, stepper, and WASM engines to guarantee all three agree.

### Source-of-truth / keep it simple
- The `.tex` tables stay the **human source of truth**.
- The committed TS table is a **faithful transcription** — explicit and in git (its whole value is being reviewable against the `.tex`). **Do not** parse `.tex` at runtime (brittle) and **do not** put AI in the CI loop (non-deterministic).
- AI is an **authoring aid**: when a `.tex` table changes, a human/AI updates the committed table in the same PR; CI then runs that committed table deterministically.
- The **implementation** is the runtime truth: the test catches impl ≠ table, and code review catches table ≠ `.tex`.

Net: one committed test, a small variant-parameterized table, a comment citing the two `.tex` files, and the cross-product loop. No `.tex` parser, no cross-table tooling, no runtime AI.

martin-henz:
Closes #198. Related: #205 (WASM follow-up), #206 (SVML follow-up).

## Part A — factored spec tables

The two operator typing tables had drifted apart by hand. They are now factored with `\input` so common rows live once:

```
python_typing.tex    = front + middle_12 + back    (Python §1/§2)
python_typing_3.tex  = front + middle_34 + back    (Python §3/§4)
```

The diff between the middle fragments *is* the documented chapter delta:

| | §1/§2 | §3/§4 |
|---|---|---|
| `==`, `!=` | numeric×numeric, string×string only | **any × any → bool** (structural on lists, CPython bool-as-int) |
| `is`, `is not` | rejected at validation | error if either operand is number/string/bool; reference identity on list/function/None |
| `>,>=,<,<=` | int/float and string×string | also admits **bool** (CPython: `True < 2`) |

Semantic decisions (Martin): §3/§4 behave like CPython on these operators — `True == 1`, `[True] == [1]`, `nan != nan` — while `is` is restricted to reference types (identity of immutables is unobservable; the interning trap), keeping `xs is None` valid.

## Part B — spec-driven conformance tests

`src/tests/operator-conformance.test.ts` transcribes the four `.tex` fragments into a committed TS table (the `.tex` stays the human source of truth; update both in the same PR; no runtime `.tex` parsing) and sweeps the full `operator × type × type` cross product at all four chapters: allowed combinations must produce the spec'd result type, everything else must raise `UnsupportedOperandTypeError`. This pins the negative space — the restrictions — and answers the issue's open question: the §1/§2 `==` restriction *is* enforced by the engine. Directed value tests cover the value-dependent cases (`**` sign, bool equality/ordering, structural list equality, None, NaN).

## Three pre-existing bugs found and fixed

1. **`x is not y` evaluated as `x is y`** — the grammar rewrote only the token's lexeme, never its type, and nothing reads the lexeme. Now emits `TokenType.ISNOT` (and the same latent twin in `not in` now emits `NOTIN`).
2. **`NoIsOperatorValidator` never fired** — `is` parses as `ExprNS.Compare`, the validator matched `ExprNS.Binary`, so chapters 1/2 silently allowed `is`. Fixed and covered by the sweep.
3. **`compile-grammar` truncated the generated parser on macOS** — BSD sed aborts with "illegal byte sequence"; the sed step now runs under `LC_ALL=C`.

## Internal review round (pre-PR)

An 8-angle review with per-candidate verification produced 10 findings, fixed here:
- `None == None` was **False** at §3/§4 (fresh none objects, reference-compared) — now True.
- Full CPython NaN semantics: NaN unequal to everything including itself, unordered; the identity shortcut still wins for the same NaN object *inside containers*, exactly CPython's per-element `is`-then-`==` rule (`[math_nan] == [math_nan]` is True, `math_nan == math_nan` is False).
- The `equal` prelude now matches `==` (bool-as-int) via an explicit coercion that stays legal at §2, where `==` takes no bool operands; `list.prelude`'s `equal` also gains the complex leaf that `linked-list`'s already had.
- `variant >= 3` gate on the runtime `is` branch (restores the §1/§2 runtime backstop), dead-code removal, helper extraction.

## Out of scope / follow-ups

- WASM (#205), SVML (#206), and Stepper (#212) engines.
- `//` and unary `+` are implemented but have no row in the spec table — excluded from the sweep and flagged in the test header for a spec ruling.
- The float lexer rejects exponent literals without a decimal point (`1e400`); Python accepts them. Noticed while writing NaN tests.
- Transcribing the book's printed `equal` definition to match, and the same bool fix in the CPython `sicp` package (`python/sicp/list.py`, PR #195).

## Test plan

- Full suite: **2,301 passing** (from 2,214 on main), including the new ~2,400-combination conformance sweep at all four chapters and directed value tests.
- `tsc --noEmit`, `prettier --check`, `eslint` clean.


src/engines/cse/operators.ts
```diff
@@ -213,12 +320,53 @@ export function evaluateBinaryExpression(
...
+  if (
+    variant >= 3 &&
+    (operator == TokenType.LESS ||
+      operator == TokenType.LESSEQUAL ||
+      operator == TokenType.GREATER ||
+      operator == TokenType.GREATEREQUAL)
+  ) {
+    left = asIntIfBool(left);
+    right = asIntIfBool(right);
+  }
```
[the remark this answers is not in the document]
martin-henz: fixed

src/engines/cse/operators.ts
```diff
@@ -120,81 +120,210 @@ export function evaluateUnaryExpression(
...
+/**
+ * Whether a value is excluded from Python §2's any x any equality: bool (avoiding
+ * CPython's bool-as-int equality, e.g. `True == 1`, as a directly written §2
+ * comparison) and function values (equality without `is` is left undefined until
+ * §3/§4 introduces it). See docs/specs/python_typing_middle_2.tex:
+ * `==,!= bool,function x any -> error`, `==,!= any x bool,function -> error`.
+ */
+function excludedFromChapter2Equality(value: Value): boolean {
+  return value.type === "bool" || value.type === "closure";
+}
```
Akshay-2007-1, line 271: excludedFromChapter2Equality excludes "bool" and "closure" but not "builtin", so at §2, comparing a builtin (e.g. head == head, head == 1) silently returns a bool instead of raising UnsupportedOperandTypeError as the spec requires. 

The conformance sweep's "function" representative is always a lambda closure, never a builtin, so this gap is completely untested.

Confirmed after testing : 

Image

The above was evaluated with Python chapter 2, it should have errored out.

src/engines/cse/operators.ts
```diff
@@ -120,81 +120,210 @@ export function evaluateUnaryExpression(
...
 
-  // Some types have value-based equality (e.g. strings), while others have reference-based equality (e.g. lists).
+  // Lists compare element-wise, recursively, as in Python
+  if (left.type == "list" && right.type == "list") {
+    return (
+      left.value.length === right.value.length &&
+      left.value.every((element, i) =>
+        structuralEquals(code, command, context, element, right.value[i]),
+      )
+    );
```
Akshay-2007-1, line 250: the §2 bool-exclusion is only checked on top-level operands before dispatching into structuralEquals; the recursive list/pair case never re-applies it and unconditionally coerces bool→int.

```python
print(pair(1, 2) == pair(True, 3))
```

Image

Recursion and nesting should be handled for chapter 2.

docs/specs/Makefile
```diff
@@ -10,7 +10,7 @@ HELPERSTEX = $(HELPERS:%=python_%.tex)
 
 all: $(PDFSPECS)
 
-%.pdf: %.tex 
+%.pdf: %.tex $(HELPERSTEX)
```
Akshay-2007-1, line 15: the new %.pdf: %.tex $(HELPERSTEX) dependency references HELPERSTEX = $(HELPERS:%=python_%.tex), but HELPERS is never defined anywhere in the file, so it expands empty and the added dependency is a silent no-op — editing the new spec fragments won't trigger a PDF rebuild via make.

src/tests/operator-conformance.test.ts
Akshay-2007-1, line 1: run() monkeypatches the shared Stash.prototype.pop on every call (~3,800 invocations across the sweep) instead of using the existing jest.spyOn(Stash.prototype, "pop") + restoreAllMocks() pattern already established in src/tests/utils.ts; besides the per-call overhead, mutating a shared prototype makes the otherwise-independent combinations unsafe to parallelize.

src/engines/cse/operators.ts
Akshay-2007-1: Another combination leading to an error

```python
print(pair(head, 2) == pair(head, 2))
```

Image

Akshay-2007-1 (changes requested):

src/engines/cse/operators.ts
martin-henz: fixed and test case added

src/engines/cse/operators.ts
martin-henz: fixed and test case added

docs/specs/Makefile
martin-henz: fixed

src/tests/operator-conformance.test.ts
martin-henz: fixed

Akshay-2007-1 (approved):
structuralEquals in src/engines/cse/operators.ts (around line 163+) recurses via plain native JS function calls when comparing nested lists/pairs element-by-element — and it has no depth counter or bound of any kind:

```ts
if (left.type == "list" && right.type == "list") {
  return (
    left.value.length === right.value.length &&
    left.value.every((element, i) =>
      structuralEquals(code, command, context, operator, element, right.value[i], restrictChapter2),
    )
  );
}
```

The interpreter does have a recursion-limit mechanism (RecursionError, a friendly RuntimeSourceError with source line/snippet/hint), but that machinery only fires from the CSE machine's main step loop, by counting instructions on the control stack as Python-level function calls happen. It has nothing to do with this native JS recursion happening entirely inside one evaluateBinaryExpression call.

So a code like this per se will fail and not give any output.

```python
x = [1]
for i in range(100):
    x = [x]
y = [1]
for i in range(100):
    y = [y]
print(x == y)
```

If this can be ignored as per the specs and the `stepLimit` in the frontend, it should be alright then!

LGTM otherwise!

PR #221, opened 2026-07-06
CSE machine: Fix Tail Call Optimisation 

AaravMalani:
In the previous TCO PR (#150), the CSE machine relied on the presence of the `RESET` instruction to detect an optimisation possibility. This PR removes the `RESET` instruction, instead using the `InstrType.ENVIRONMENT` instruction to detect tail calls.


src/engines/cse/interpreter.ts
```diff
@@ -1152,10 +1137,12 @@ const cmdEvaluators: CmdEvaluators = {
     if (
       topElement !== undefined &&
       isInstr(topElement) &&
-      topElement.instrType === InstrType.RESET
+      topElement.instrType === InstrType.ENVIRONMENT
```
martin-henz, line 1169: Hmm, I don't understand. In case of a tail call, you propose to pop the existing env instruction and replace it with the environment of the caller? Should we not rather keep the existing env instruction on the control and discard the environment of the caller?

martin-henz (changes requested):

src/engines/cse/interpreter.ts
AaravMalani: We should keep the existing env instruction on the control, but it also makes sense to clean up the environment array.

When a TCO possibility is detected, it restores the environment array to the environment the parent function's callee

AKA,
```
def f(x):
    def g(x):
        return f(x - 1)
    return g(x)
f(2)
```

When the application instruction for g(2) is being processed, it restores the environment to f's parent environment (aka the program environment). Then, it pushes the new environment into the environments array and pushes an identical environment instruction for the parent environment

I'll make an optimisation so that it doesn't push and pop the same instruction if that helps clarify

martin-henz: OK. This is better: More efficient (I think) and easier to understand.

martin-henz (approved):
OK!

PR #226, opened 2026-07-07
PVML: fix compiler bugs, wire native Pynter groups/prelude, correct primitive mappings, add operator sweep

martin-henz:
## Summary

Consolidates a session's worth of native-Pynter parity work into one PR. Started from investigating why `yarn pynter:report` (added in #220) showed low pass rates and worked backwards to root causes — most of what's here turned out to be real, independent bugs rather than missing features.

- **Fix PVML compiler bugs in recursion, control flow, and list literals** — four bugs found by chasing native-Pynter failures that also reproduced on the pure-TS `PVMLInterpreter` (ruling out anything Pynter-specific): slot/env-builder bookkeeping wasn't shared across a compilation (a function calling a sibling or itself recursively could silently invoke the wrong closure at the wrong arity), falling off the end of a function body leaked the last statement's value instead of yielding `None`, subscript-target assignment (`xs[i] = v`) crashed the compiler outright, and list literals pushed a size operand `NEWA` never consumes, corrupting stack balance for the rest of the program (eventually manifesting as a native stack overflow).
- **Wire canonical per-variant groups and prelude compilation into native Pynter** — `runCodePvmlDetailed` was hardcoded to `[misc, math]` regardless of variant and never compiled any group's prelude, so linked-list/list/pairmutator/stream programs always failed before reaching native Pynter at all.
- **Correct and extend PVML's native primitive index mappings** — `PRIMITIVE_FUNCTIONS` had several indices that didn't match native Pynter's actual dispatch table (`abs`/`min`/`max`/`round`/`len` were silently calling `error()`/`is_null()`/`is_number()`/`length()`/`map()` instead), plus wires up ~40 more primitives with ready native equivalents.
- **Add a native Pynter operator-conformance sweep** — extends #207's CSE operator-conformance sweep to native Pynter, computing the CSE machine's own result live as the expected value for every `operator × type × type` combination at all four chapters, rather than a second hand-written table that could drift.

Net result on `yarn pynter:report` (opt-in, requires `PYNTER_RUNNER_PATH`): total pass rate roughly 46% → 78% across existing suites, `stream` 56%→100%, `linked-list` 45%→91%, `list` 44%→94%, plus the new operator sweep at 92% (2596/2812).

## Dependency note

Five new primitive names (`is_integer`, `is_float`, `is_complex`, `_gen_list`, `arity`) are mapped to indices 92–96, which don't exist in native Pynter today — they're being added in a companion PR to the [pynter](https://github.com/source-academy/pynter) repo, landing separately. This PR doesn't depend on that landing first: all native-Pynter tests are gated behind `PYNTER_RUNNER_PATH` and skip entirely without it, so CI is unaffected either way. Against a stock (pre-companion-PR) Pynter binary, only calls to those five primitives would fault; everything else in this PR is verified against Pynter as it exists upstream today.

## Test plan

- `npx tsc --noEmit`, `yarn lint`, `yarn format:ci` all clean.
- Full suite: 2495 passing (up from 2401 on main), 0 failures.
- Native Pynter suites (`PYNTER_RUNNER_PATH=<runner binary>`): verified suite-by-suite during development; no regressions against a fresh run.

### DIFF
--- a/scripts/pynter-parity-report.ts
+++ b/scripts/pynter-parity-report.ts
@@ -129,7 +129,7 @@ function main() {
     return [suite, `${passed}/${attempted}`, String(skipped), pct(passed, attempted)];
   };
 
-  const header = ["Suite", "Pass/Attempted", "Skipped (complex numbers)", "Pass rate"];
+  const header = ["Suite", "Pass/Attempted", "Skipped (out of scope)", "Pass rate"];
   const body = rows.map(r => toRow(r.suite, r.passed, r.failed, r.skipped));
   body.push(
     toRow("**Total**", totals.passed, totals.failed, totals.skipped).map((cell, i) =>
--- a/src/engines/pvml/PVMLIRBuilder.ts
+++ b/src/engines/pvml/PVMLIRBuilder.ts
@@ -270,7 +270,7 @@ const STACK_EFFECTS = new Int16Array(OPCODE_MAX + 1);
   STACK_EFFECTS[OpCodes.STPB] = -1;
 
   // Array operations
-  STACK_EFFECTS[OpCodes.NEWA] = 0; // Takes size, produces array
+  STACK_EFFECTS[OpCodes.NEWA] = 1; // Takes no operand, produces an empty array (see visitListExpr)
   STACK_EFFECTS[OpCodes.LDAG] = -1;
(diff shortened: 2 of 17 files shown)


martin-henz:
coverage decreases because the new operator tests are not activated yet

PR #227, opened 2026-07-07
CSE: Fix nested lambdas 

### LINKED ISSUE #222: CSE Evaluator: Nested Lambdas raise an error
## Sample
```py
print(lambda t : lambda u : t + u)
```

## Expected Output
```
<function (anonymous)>
```

## Actual Output
```
t.map is not a function
```

AaravMalani:
This PR fixes #222, by allowing the body in the CSE machine snapshot collector to wrap itself in an array, so that `map` would work on the resultant body

### DIFF
--- a/src/conductor/plugins/PyCseMachinePlugin.ts
+++ b/src/conductor/plugins/PyCseMachinePlugin.ts
@@ -303,15 +303,19 @@ function serializeControlItem(item: ControlStackItem, code: string): SerializedI
     // For block-like nodes pass body length and child types so the adapter can build
     // stub body arrays (used by ControlExpansionAnimation / StatementSequence handling).
     if (
-      jsNodeType === "StatementSequence" ||
-      jsNodeType === "FunctionDeclaration" ||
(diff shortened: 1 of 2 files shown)


Akshay-2007-1 (approved):
LGTM! I tested it locally and stress-tested with multiply nested lambdas. Thanks for the update!

One non-blocking nit: the body-unwrap fallback now silently omits bodyLength/bodyNodeTypes when item.body is undefined (previously emitted bodyLength: 0). Confirmed this only affects the root FileInput control item (which never carried a real .body property to begin with) and has no visible effect on the visualizer, not worth blocking on, just flagging in case it matters for other consumers of this metadata down the line. I did test it and there isnt a correctness issue per se YET.

AaravMalani:
> ned (previously emitted bodyLength: 0). Confirmed this only affects the root FileInput control item (which never carried a real .body property to begin with

Yep, I noticed that there will always be a `.body` property for the given types, so it's just for extra typing safety

PR #230, opened 2026-07-07
Fix CSE Machine visualizer display bugs: position-0 node mislabeling, verbose builtin display

### LINKED ISSUE #228: CSE Machine: first token of the source is mis-rendered on the control stack (generic label, no line info)
## Summary

In the CSE Machine visualizer, whichever AST node's `startToken` happens to be the very first token of the whole source file (source index 0) gets mis-serialized on the control stack: it loses its real display text (falls back to a generic per-kind label) and always loses `startLine`/`endLine` metadata. This is most visible with a call like `print(1)` as line 1: at the step where the callee is pushed onto control, it renders as `"var"` instead of `"print"`. Move the exact same code down to line 2 (e.g. prepend a blank line) and it correctly renders as `"print"` with proper line info.

## Repro

Using `collectSnapshots` / `runCode` directly (variant 1, `misc` + `math` stdlib groups):

```python
print(1)
```
vs
```python

print(1)
```

Stepping both and comparing `control[0]` (top of control stack) at each step:

**`print(1)` on line 1** — step 2:
```json
{"displayText":"var","metadata":{"nodeType":"Identifier"}}
```

**`print(1)` on line 2 (blank line 1)** — step 2:
```json
{"displayText":"print","metadata":{"startLine":2,"endLine":2,"nodeType":"Identifier"}}
```

Same code, same relative step, different (and wrong, in the line-1 case) output. This isn't specific to `print` or to `Call` callees — it reproduces for *any* single-token leaf node that lands at source index 0. E.g. a bare literal as the entire program:

```python
42
```
Step 0 (the wrapping `ExpressionStatement`) renders as `"expr"` (generic fallback) instead of `"42"`, with no line metadata. Step 1 (the `Literal`/`BigIntLiteral` node itself) happens to still show `"42"` (coincidentally, via the separate synthetic-BigIntLiteral fallback branch — see root cause), but likewise has no line metadata.

## Root cause

`src/conductor/plugins/PyCseMachinePlugin.ts`, `serializeControlItem` (around line 262-289):

```ts
const start: number = item.startToken?.indexInSource ?? -1;
...
const isRealSourceNode = start > 0 || (endTok?.indexInSource ?? 0) > 0;
```

This guard exists to detect *synthetic* nodes created at runtime (e.g. implicit `range()` start/step bounds for `for` loops) whose tokens are deliberately pinned to `indexInSource: 0` — see `src/engines/cse/utils.ts` around lines 698-728 (`tempTokenZero`, `tempTokenOne`, `literalToken`, all constructed as `new Token(..., 0, 0, ...)`).

The problem: a **real** node can also legitimately have `indexInSource === 0` — namely, whatever token happens to be the very first token of the actual source file. For a single-token node (an `Identifier`/`Variable`, or a bare `Literal`/`BigIntLiteral`) at that position, both its `startToken` and `endToken` are literally the same token, so both `start > 0` and `endTok.indexInSource > 0` are false, and the guard incorrectly classifies it as synthetic. It then falls into the `else` branch and returns `KIND_LABELS[item.kind]` (e.g. `"var"` for `Variable`, `"expr"` for `SimpleExpr`) instead of slicing the real source text, and skips setting `loc`/`startLine`/`endLine` entirely.

Multi-token nodes (e.g. the whole `print(1)` `CallExpression`, or `print(1)` as a statement) aren't affected because their `endToken` is a later token with `indexInSource > 0`, so the guard passes even though `start === 0`.

## Suggested direction (not prescriptive)

The guard is trying to distinguish "synthetic runtime-generated node" from "real node at position 0" using position arithmetic, but position 0 is ambiguous between the two. A more robust fix would tag synthetic tokens/nodes explicitly (e.g. a `synthetic: true` marker set when constructing `tempTokenZero`/`tempTokenOne`/`literalToken` in `utils.ts`) and have `serializeControlItem` check that flag instead of inferring it from `indexInSource`. That would let real position-0 nodes render correctly while still filtering out genuinely synthetic ones.

## Impact

- Any program whose very first statement begins with a single-token leaf (a bare name, call to a builtin/function with no preceding whitespace, bare literal, etc.) will show incorrect/generic labels and missing line metadata for that node in the CSE Machine control-stack visualization the first time it's pushed/evaluated. Since almost every SICPy program's first line is exactly this shape, this is a very common visual glitch, not an edge case.

### LINKED ISSUE #229: CSE Machine: builtin values render as verbose `<built-in function xxx>` on the stash instead of a compact chip
## Summary

When a builtin (e.g. `print`, `len`, a `math_*`/stdlib function, etc.) ends up on the stash in the CSE Machine visualizer, it's displayed as the full Python-`repr()`-style string `<built-in function xxx>`. That's correct for actual Python `repr()`/`print()` output, but it's long and visually noisy as a stash "chip" compared to how every other callable value is shown there.

For comparison, user-defined closures are already rendered compactly: `formatValue()`'s `"closure"` case returns just `cl.node.name.lexeme` (e.g. `"my_func"`) or `"lambda"` — no `<function ...>` wrapper — with the "this is a function" fact conveyed separately via `label`/`metadata`, not baked into the display string.

## Where this comes from

`src/conductor/plugins/PyCseMachinePlugin.ts`, `formatValue()`:

```ts
case "closure": {
  const cl: Closure = v.closure;
  return cl.node.kind === "FunctionDef" ? cl.node.name.lexeme : "lambda";
}
case "function":
  return v.name || "function";
...
case "builtin":
  return \`<built-in function \${v.name}>\`;
```

The `"builtin"` branch is the odd one out — it's the only callable-value case that wraps the name in Python-repr syntax instead of just showing the name. (This same `<built-in function ...>` string template is also used, separately and correctly, in `src/utils/stringify.ts` and `src/stdlib/utils.ts` for actual Python `repr()`/`print()` semantics — those should be left alone; this issue is only about the CSE Machine's own stash serialization in `PyCseMachinePlugin.ts`.)

Relatedly, `serializeValue()` sets `label: typeTranslator(v.type)` for every value, but `typeTranslator()` (`src/engines/cse/types.ts`) has no case for `"builtin"` (or `"function"`), so it falls through to `"unknown"`:

```ts
export function typeTranslator(type: Value["type"]): string {
  switch (type) {
    case "bigint": return "int";
    ...
    case "closure": return "function";
    default: return "unknown";
  }
}
```

So today a builtin's `label` is literally `"unknown"` while its `displayValue` carries all the "this is a builtin function" information baked into a long string — the opposite of how closures do it.

## Ask

Tidy up the builtin display to match the closure convention, as a non-breaking change scoped to `PyCseMachinePlugin.ts` (i.e. only the debugger/visualizer serialization — not `stringify.ts` / `stdlib/utils.ts`, which must keep producing real CPython-parity `repr()` output):

1. `formatValue()`'s `"builtin"` case should return just `v.name` (e.g. `"print"`), not the wrapped `<built-in function print>` string.
2. Add a `"builtin"` case to `typeTranslator()` so `label` reflects it distinctly from closures — CPython itself distinguishes these via `type(print).__name__` (`"builtin_function_or_method"`) vs `type(my_func).__name__` (`"function"`), so the label could mirror that instead of falling through to `"unknown"`.
3. Longer term / separate consideration: right now `BuiltinValue` (`src/engines/cse/stash.ts`) only carries a flat `name` — there's no record of which stdlib group a builtin came from (core `misc`/`math` vs `linked-list`/`list`/`pairmutator`/`stream`/`parser`, registered per-variant in `src/runner.ts`'s `VARIANT_GROUPS` loop). If we want the visualizer to eventually distinguish "core builtin" vs "stdlib function" (as opposed to just "builtin" vs "user-defined closure"), that group name would need to be threaded through at registration time and carried on `BuiltinValue`/into `serializeValue()`'s metadata. Flagging this as a possible follow-up rather than part of the minimal fix, since it touches the `Value` type shape rather than being purely a serialization-layer change.

## Non-goals

- Do not change `formatValue`/`repr()` used for actual `print()`/`repr()` program output (`src/utils/stringify.ts`, `src/stdlib/utils.ts`) — those must stay CPython-accurate.
- Should not change the `Value`/`BuiltinValue` runtime type shape for the minimal fix (items 1-2 above) — only the CSE Machine's own display serialization.

Akshay-2007-1:
## Summary

Fixes two CSE Machine visualizer display bugs, both scoped to the serialization layer (`src/conductor/plugins/PyCseMachinePlugin.ts` and friends) — no change to actual evaluation semantics or `repr()`/`print()` output.

Closes #228, closes #229.

### #228 — real nodes at source position 0 mislabeled as synthetic

`serializeControlItem` inferred "this is a synthetic runtime-generated node" (e.g. implicit `range()` start/step bounds) from `startToken.indexInSource === 0`. But a real node whose token is literally the first token of the source also has index 0, so any single-token leaf (a bare `Identifier`, a bare `Literal`) at that position was misclassified: it rendered with a generic `KIND_LABELS` fallback and no line metadata instead of its real text.

Concretely: `print(1)` as the very first line renders its callee as `"var"` instead of `"print"` at the control-stack step where it's pushed; the identical code moved to line 2 renders correctly.

Fix: mark runtime-fabricated tokens with an explicit `synthetic` flag on `Token` (`src/tokenizer/tokenizer.ts`), set it at all the call sites that build synthetic tokens (`src/engines/cse/utils.ts`, `src/engines/cse/interpreter.ts`), and check that flag in `serializeControlItem` instead of inferring syntheticness from position.

### #229 — builtins rendered with a verbose `repr()`-style string

`formatValue()`'s `"builtin"` case returned `` `<built-in function ${name}>` ``, the same string used for real Python `repr()` output elsewhere in the codebase. Every other callable value in the visualizer is shown compactly instead — closures render as just their name (`funcName` / `"lambda"`), with "this is a function" carried via `label`/metadata, not the display string.

Fix: `formatValue()`'s `"builtin"` case now returns just the name, and `typeTranslator()` gains a `"builtin"` → `"builtin_function_or_method"` case (matching CPython's `type(print).__name__`) so `label` still distinguishes builtins from user-defined closures.

`src/utils/stringify.ts` and `src/stdlib/utils.ts`, which produce actual Python `repr()`/`print()` output, are untouched — this PR only touches the debugger/visualizer's own serialization.

## Test plan

- Added regression test coverage for both fixes directly in `src/tests/PyCseMachinePlugin.test.ts` (in the existing `serializeControlItem` / `formatValue` / `serializeValue` describe blocks), matching this repo's one-file-per-module test convention.
- Updated two assertions in that file that asserted the old `<built-in function xxx>` string, since #229 makes them factually wrong.
- `npx tsc --noEmit` clean.
- `npx eslint` clean on all changed files.
- Full `npx jest` run: all suites pass (confirmed after an earlier false claim in this description about a pre-existing compile error — that was a stale local `node_modules` on my end, not a real repo issue; thanks @Shrey-D for catching it).
- Manually verified via a repro script stepping the CSE machine over `print(1)` on line 1 vs. line 2, and over a `for x in range(3)` loop to confirm synthetic range bounds still render correctly (as bare numeric values, unaffected by this change).


Shrey5132:
## Review

### Overview

This fixes two display-only bugs in the CSE Machine visualizer (`src/conductor/plugins/PyCseMachinePlugin.ts`), closing #228 and #229:

- **#228**: A real AST node whose token happens to sit at source index 0 (e.g. `print` as the first token of the file) was indistinguishable from a runtime-fabricated node (e.g. implicit `range()` bounds), which also gets pinned to index 0. Both were detected via `indexInSource === 0`, so real position-0 nodes rendered with a generic fallback label instead of their real text.
- **#229**: Builtins rendered as `<built-in function xxx>` (repr-style) in the stash, inconsistent with closures, which render as just their name.

The fix adds an explicit `synthetic?: boolean` flag to `Token` (`src/tokenizer/tokenizer.ts:135`), sets it at all runtime call sites that fabricate tokens (`interpreter.ts`, `utils.ts`), and switches `serializeControlItem`'s reality check from position-inference to reading that flag (`PyCseMachinePlugin.ts:278`). Separately, `formatValue`'s `"builtin"` case now returns just the name, and `typeTranslator` gains a `"builtin"` → `"builtin_function_or_method"` case so the label still carries the distinction.

### Verification performed

Checked out the branch and verified independently rather than just trusting the description:

- `npx tsc --noEmit` — clean.
- `npx eslint` on all changed files — clean (one pre-existing `no-unused-eslint-disable` warning in `PyCseMachinePlugin.test.ts`, confirmed present on `main` too — unrelated to this PR).
- Full `npx jest` on this branch: **26/26 suites, 2493/2493 non-skipped tests pass.**
- Full `npx jest` on `main` for comparison: **24/24 suites, 2487/2487 pass.** The delta is exactly the 2 new test files / 6 new tests — no regressions.
- Traced all 8 call sites in the codebase that construct `ExprNS`/`StmtNS` nodes with explicit tokens to confirm start/end token pairs are always same-origin (both synthetic or both real) — so the new `!start.synthetic && !end.synthetic` check is behaviorally sound; there's no existing mixed-origin node it would mishandle.
- Checked `src/resolver/resolver.ts`'s sentinel tokens (also constructed at index 0, e.g. lines 17, 220–227, 376, 439) — confirmed these only feed the resolver's static `Environment.names` table and never reach the CSE machine's `control`/`stash`, so they're correctly out of scope here.
- Confirmed the two review-bot comments on this PR (duplicate `typeTranslator` in `utils.ts`; wrong `TokenType.NUMBER` on synthetic BigInt tokens) were both already addressed in follow-up commit `88a8f90`.

### One thing worth fixing: a claim in the description doesn't hold up

The description justifies splitting new tests into two new files instead of extending `src/tests/PyCseMachinePlugin.test.ts` by saying that file "currently has a pre-existing, unrelated compile error — `globalNames` missing on `CseSerializedEnvFrame` — that prevents the whole file from running; confirmed this fails identically on a clean checkout of main."

I couldn't reproduce this:
- `globalNames?: string[]` is present on `CseSerializedEnvFrame` (`node_modules/@sourceacademy/common-cse-machine/dist/index.d.ts:100`, added back in #196).
- Running `PyCseMachinePlugin.test.ts` directly on a clean `main` checkout (`3e37c71`) passes all 78 tests, including `"carries globalNames from the closure's globalVariables"`.
- The full-suite run on `main` above shows zero failing suites anywhere in the repo.

So the file split doesn't seem to be blocked by anything. The two assertion updates in `PyCseMachinePlugin.test.ts` (the ones asserting the old `<built-in function xxx>` string) are correct and necessary on their own — but the new #228/#229 coverage could just as well live in that same file, which would also match this repo's existing convention (every other file in `src/tests/` is a single simple name, e.g. `lexer.test.ts`, `resolver.test.ts`).

Could you either fix the description or fold the two new test files into `PyCseMachinePlugin.test.ts`? Mostly flagging so the "pre-existing compile error" claim doesn't mislead someone later.

### Minor, non-blocking

`currentLine` in `PyCseMachinePlugin.ts` (~line 460) reads `context.runtime.nodes[0]?.startToken?.line` without a synthetic check. Synthetic tokens are constructed with `line: 0`, so if a synthetic node ever becomes `nodes[0]`, the frontend would get `currentLine: 0`. This is pre-existing and not touched by this PR — just flagging as a possible follow-up.

### Overall

Low risk. The change is narrowly scoped to the debugger/visualizer serialization layer with no path to affecting evaluation semantics, and is covered by targeted regression tests for both issues. Type-check, lint, and full test suite are all green with no regressions from my independent run.

Shrey5132 (approved):

Akshay-2007-1:
@Shrey5132 you're right, thanks for catching it — my node_modules was stale (`@sourceacademy/common-cse-machine@0.1.0` installed instead of the `yarn.lock`-pinned `0.2.0`, which is where `globalNames` was added back). Ran `yarn install` and confirmed `PyCseMachinePlugin.test.ts` compiles and passes cleanly on its own.

Folded the #228/#229 coverage into that file's existing `serializeControlItem`/`formatValue`/`serializeValue` describe blocks and deleted the two standalone files in 2d8666f, and corrected the PR description's test-plan section to drop the false claim.

On the non-blocking `currentLine`/synthetic-token nit — agreed it's a real gap, but leaving it out of this PR since it's pre-existing and unrelated to what #228/#229 touch. Happy to pick it up separately if useful.

Akshay-2007-1:
@Shrey5132 filed your \`currentLine\`/synthetic-token nit as #233 — turns out it's more than theoretical. I stepped `for x in range(3): print(x)` through `collectSnapshots` directly and confirmed `currentLine` never returns to line 1 (the `for` line) after the very first step; every loop-continuation phase (condition re-check, increment) collapses it to `0` for the rest of execution, repeating every iteration. Full trace is in the issue. Keeping it out of this PR as agreed, but wanted to confirm on-record that it's a real, reproducible bug and not just a hypothetical edge case.

PR #231, opened 2026-07-08
Revamping equality and identity

### LINKED ISSUE #206: Bring the SVML engine in line with the revised operator semantics

### LINKED ISSUE #212: Bring the stepper in line with the revised §1/2 operator semantics
Follow-up to #198 (sibling issues for the other engines: #205 WASM, #206 SVML). The `operators` branch (#207) revises the operator typing spec and implements it in the **CSE engine** (`src/engines/cse/operators.ts`), with a spec-driven conformance sweep (`src/tests/operator-conformance.test.ts`). The **stepper** (`src/conductor/stepper/`, a substitution-based evaluator, separate from the CSE engine) was not touched by #207 — it needs the same semantics, and currently does not have them.

## Current stepper behavior (before this issue)

`src/conductor/stepper/reduce.ts`'s binary-operator reduction (`contractBinary`, `reportBinaryTypeError`, `valueTypeName`, `stringBinary`) has **no chapter/variant parameter at all** — it implements a single, chapter-blind rule set that predates #207's spec revision:
- `==`/`!=` only for (numeric, numeric) or (string, string) — everything else is a `TypeError` if both operands are concrete values (e.g. `True == 1`, `None == None`), or a silent "stuck" if a pair/list is involved (equality on `ArrayExpression` nodes was never modeled: `valueTypeName` deliberately returns `null` for them so no error is even reported).
- `bool` is excluded from `isNumericValue` unconditionally, so it never participates in equality or ordering, at any chapter.
- `is`/`is not` are rejected up front by `preprocessPython` (`src/conductor/stepper/preprocess.ts`) as operators the substitution model can't represent — a separate, pre-existing, deliberate design choice that is *not* part of this issue's scope (the stepper was never meant to model `is`).

Chapter-awareness *does* already exist in the stepper for name resolution and builtin availability — `preprocessPython(fileInput, source, chapter)` delegates feature-gating to the same `analyze()`/`Resolver` pipeline the CSE engine uses, and `getAvailableBuiltinNames(chapter)` / `isBuiltinFunctionAvailable(name, chapter)` correctly gate e.g. `pair` to chapter 2+. It simply was never threaded into the operator-reduction path.

## What changed in the CSE engine (the reference behavior, see #205/#206 for the full per-chapter breakdown)

- **§1**: unchanged — `==`/`!=` only numeric × numeric and string × string; ordering only int/float and string × string; `is`/`is not` rejected at validation. This matches the stepper's current (accidental) behavior, so §1 is not affected by this issue.
- **§2**: now **structural over any × any except bool and function**, which are excluded entirely (`bool,function x any -> error`) — cross-type comparisons (`1 == 'abc'` → `False`, not an error) and pairs/None (`pair(1,2) == pair(1,2)`, `pair(1,2) == None` → `False`) all participate. Ordering unchanged from §1 (still int/float only, no bool). This is a real divergence from the stepper's current behavior, which still treats §2 identically to §1.

So merging #207 as-is makes the stepper diverge from the CSE engine at §1 **and** §2 simultaneously — not just "not yet updated," but actively producing wrong results or hanging on previously-uncommon-but-now-valid programs.

## Work

- [ ] Thread a `chapter`/`variant` parameter into the stepper's binary-operator reduction path (`contractBinary` and friends in `reduce.ts`), the way it's already threaded into `preprocessPython`/`getAvailableBuiltinNames`.
- [ ] Reimplement the §1/§2 semantics above in `StepNode`/`ArrayExpression` terms (the stepper's data representation is a plain AST-node shape, not the CSE engine's tagged `Value` union, so this is a parallel reimplementation, not shared code — same situation as WASM/SVML).
- [ ] Model pair/list structural equality (`ArrayExpression` operands), which the stepper has never reduced at all.
- [ ] Extend the conformance sweep in `operator-conformance.test.ts` to run its table against the stepper too (part of the multi-engine stretch goal in #198), so it cannot drift from the spec or the CSE engine again.

martin-henz:
This PR adjusts the meaning of ==, !=, **is,** and **is not** in all Python §x variants and in all implementations, according to https://github.com/source-academy/py-slang/issues/232. This affects the stepper, CSE machine, and PVML compiler, and many test cases.

Current status:
* PVML compiler: done
* CSE machine: done
* Stepper: done

Fixes #212 
Fixes #206

Ready for review!

## Summary by CodeRabbit

* **New Features**
  * Added support for Python identity operators (`is` / `is not`) alongside improved equality behavior.
  * Comparison results now better match Python-like structural and identity semantics for values such as pairs, `None`, numbers, strings, and booleans.

* **Bug Fixes**
  * Improved handling of invalid comparisons involving booleans and functions.
  * Updated execution to consistently use the supported Python target version.

* **Documentation**
  * Refreshed language specification notes to match the updated comparison rules.


docs/specs/python_typing_tail_34.tex
```diff
@@ -3,18 +3,18 @@
...
-    by Python 3.14).
-
+    \href{https://docs.python.org/3.14}{\color{DarkBlue}Python 3.14 Documentation}.
     As in Python, \texttt{\textbf{True}} and \texttt{\textbf{False}} are
     represented by the integers 1 and 0, respectively, for the purpose of
-    any comparisons.
+    any comparisons. The operators \texttt{\textbf{is}} and \texttt{\textbf{is not}}
+    test for identity. As in Python, we can assume that there exists only one
+    \texttt{\textbf{None}}, \texttt{\textbf{True}}, \texttt{\textbf{False}} value
+    system-wide, and that all lists and functions have their
```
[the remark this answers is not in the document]
martin-henz: yes, correct. wording improved.

src/pvml-runner.ts
```diff
@@ -61,6 +61,14 @@ export async function runCodePvmlDetailed(
...
+  // Pynter's target is Python (SICPy) §3 specifically (see pynter/README.md) —
+  // it implements that chapter's semantics unconditionally, with no runtime
+  // notion of "chapter" to gate narrower/wider rules for §1/§2/§4. Reject
+  // anything else here rather than silently running it with the wrong rules.
+  if (variant !== 3) {
+    throw new RunError("parse", `Pynter only supports Python §3; got variant ${variant}.`);
+  }
+
   const { pynterPath, groups = VARIANT_GROUPS[variant] } = options;
   if (!groups) throw new RunError("parse", `Invalid variant: ${variant}. Expected 1–4.`);
```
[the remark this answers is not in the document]
martin-henz: let's play defensive here; in future, more variants may be available for pynter

src/conductor/stepper/reduce.ts
```diff
@@ -359,6 +362,53 @@ function stringBinary(op: string, l: string, r: string): StepNode | null {
...
+  if (isPairNode(left) && isPairNode(right)) {
+    const le = left.elements as StepNode[];
+    const re = right.elements as StepNode[];
+    return structuralEquals(op, le[0], re[0]) && structuralEquals(op, le[1], re[1]);
+  }
+  if (isPairNode(left) || isPairNode(right)) return false;
+  if (isEmptyList(left) || isEmptyList(right)) return isEmptyList(left) && isEmptyList(right);
+  if (left.type !== "Literal" || right.type !== "Literal") return false;
+  const l = left.value;
+  const r = right.value;
```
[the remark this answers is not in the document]
martin-henz: Finding 1 — src/conductor/stepper/reduce.ts:400, "high priority": claims None == None falls through structuralEquals and incorrectly returns false.
False positive. Line 397 already handles this: if (isEmptyList(left) || isEmptyList(right)) return isEmptyList(left) && isEmptyList(right); — None is represented as isEmptyList, so None == None returns true before ever reaching line 400. I re-ran the exact test ("structural over any x any", which includes None == None → True) and it passes. Gemini's suggested fix (if (l === r) return true on the .value primitives) is also redundant/wrong, since null === null in JS is already true — it wouldn't even be needed if the bug were real.

martin-henz:
fixes https://github.com/source-academy/py-slang/issues/212

martin-henz:
fixes https://github.com/source-academy/py-slang/issues/206

src/engines/cse/operators.ts
```diff
@@ -308,37 +308,21 @@ export function handleExpandedEquality(
...
+ * of structure observable — and `None is None` is true. Numbers, strings and
+ * booleans have no real object identity in this interpreter (each is a freshly
+ * wrapped primitive, not a boxed object), so — as whether `1 is 1` holds is
+ * unspecified in Python anyway (interning) — they compare by their underlying
+ * value instead, the same simplification structuralEquals already makes for
+ * these types. Values of different types are never identical (so `xs is None`
+ * is simply false for a list `xs`).
  */
 function pyIdentical(left: Value, right: Value): boolean {
   if (left.type !== right.type) {
```
Akshay-2007-1, line 347:
```suggestion
  if (left === right) {
    return true;
  }
  if (left.type !== right.type) {
```

Currently the check for equality of exact same object in memory as well is incorrect : 
```python
x = math_nan
print(x is x)        # False — same object bound to itself, should be True
print(-0.0 is 0.0)   # True  — two distinct float literals, conflated by value-equality
```
The simple fix is to initially also check if they are exactly the same objects like we do for literals anyways!

Akshay-2007-1: This is the same case even for Python chapter 1 and 2 where we dont use `is` and instead the `==`

```python
x = math_nan
print(x == x) # False
print(-0.0 == 0.0) # True
```

Akshay-2007-1: Another issue I found while testing is the following code in py-slang, 
```python
x = complex(math_nan, 0)
print(x is x)
print(x == x)
```

produces the below result
```python
False 
True
```

---

A similar code in Python, say,

```python
import cmath
import math

x = complex(math.nan, 0)

print(x is x)
print(x == x)
```

produces the below result which is exactly inverted
```python
True
False
```

Akshay-2007-1 (changes requested):
There is one minor one line change that I suggested which fixes several correctness issues as per the specs!

src/engines/cse/operators.ts
martin-henz: > Another issue I found while testing is the following code in py-slang,
> 
> ```python
> x = complex(math_nan, 0)
> print(x is x)
> print(x == x)
> ```
> 
> produces the below result
> 
> ```python
> False 
> True
> ```
> 
> A similar code in Python, say,
> 
> ```python
> import cmath
> import math
> 
> x = complex(math.nan, 0)
> 
> print(x is x)
> print(x == x)
> ```
> 
> produces the below result which is exactly inverted
> 
> ```python
> True
> False
> ```

The Python version (believe it or not) correct: x is idential to itself but not structurally equal to itself. Equality follows IEEE 754, and according to the standard, NaN is not equal to anything, not even itself.

Akshay-2007-1: 😭😭
I am still getting used to these nitty gritties of Python. But even with whats being said here, we still have conflicting behaviour in py-slang vs CPython. The outputs are reversed.

martin-henz:
Fixed both in da2cd26:

- **\`is\` identity** (@Akshay-2007-1): \`pyIdentical\` now checks \`left === right\` first, before any type/value comparison — exactly your suggested fix. \`math_nan is math_nan\` and \`x = complex(math_nan, 0); x is x\` are now \`True\`, matching CPython. Added regression tests.
- **Missing \`parser\` group** (@coderabbitai): \`generateNativePynterTestCases(applyTests, 3)\` now passes \`groups\` explicitly, so \`apply_in_underlying_python\` is actually covered by that suite.

One more thing I found while verifying — flagging separately rather than folding into this fix since it's a different code path: \`complex(math_nan, 0) == complex(math_nan, 0)\` (same object) still incorrectly returns \`True\`. The top-level NaN guard in \`handleExpandedEquality\` (the one that correctly makes \`math_nan == math_nan\` → \`False\`) only checks \`value.type === "number"\` (\`isNaNValue\`), not a complex value with a NaN component, so a NaN-containing complex falls through to \`structuralEquals\`'s identity shortcut and wins incorrectly — same root cause class Akshay found, just on the \`==\` side instead of \`is\`. Happy to fix this too if wanted, just didn't want to silently expand scope on this thread.

martin-henz:
Fixed the complex-NaN \`==\` bug too, in a444d04. Root cause: \`isNaNValue\` only checked \`type === "number"\`, so a complex value with a NaN real or imaginary component never tripped the top-level "NaN is unequal to everything" guard and fell through to \`structuralEquals\`'s identity shortcut instead. CPython's NaN propagation is component-wise into complex numbers, so extended the check there too.

\`x = complex(math_nan, 0); x == x\` is now \`False\` (matching CPython), while \`x is x\` stays \`True\` (from the earlier fix) — and the container-identity-shortcut case is preserved: \`c = complex(math_nan, 0); [c] == [c]\` is still \`True\`, same as the existing \`math_nan\` behavior, since container comparison checks identity per element first. Added regression tests for all of these.

PR #234, opened 2026-07-08
Fix currentLine collapsing to 0 during for/range() loop synthetic steps

### LINKED ISSUE #233: CSE Machine: currentLine collapses to 0 during for/range() loops — the for-line's highlight never returns after first entry
## Summary

Follow-up from [PR #230](https://github.com/source-academy/py-slang/pull/230), flagged as a non-blocking nit by @Shrey5132 during review:

> `currentLine` in `PyCseMachinePlugin.ts` (~line 460) reads `context.runtime.nodes[0]?.startToken?.line` without a synthetic check. Synthetic tokens are constructed with `line: 0`, so if a synthetic node ever becomes `nodes[0]`, the frontend would get `currentLine: 0`. This is pre-existing and not touched by that PR — just flagging as a possible follow-up.

I verified this is a real, frequently user-visible bug, not just a theoretical gap — see below.

## Root cause

`src/conductor/plugins/PyCseMachinePlugin.ts:459-460`:

```ts
const currentNode = context.runtime.nodes[0] as { startToken?: { line?: number } } | undefined;
const currentLine: number | undefined = currentNode?.startToken?.line;
```

This mirrors the non-conductor CSE machine's `updateInspector`, which drives the editor's blue "current line" highlight. It doesn't check `currentNode.startToken.synthetic` (the flag added in #230). Every runtime-fabricated node used by `for`/`range()` loops — the implicit `range()` start/step bounds (`utils.ts`'s `evaluateForIterator`) and the per-iteration loop-variable increment (`utils.ts`'s `generateForIncrement`, plus the synthetic `BigIntLiteral`/`Binary` nodes built in `interpreter.ts`'s `FOR` instruction handler) — carries a token with `line: 0`. Whenever one of these becomes `nodes[0]`, `currentLine` reports `0` instead of the real source line, and stays wrong until a real node is evaluated again.

## Why this isn't a corner case

`generateForIncrement` runs on **every iteration of every `for`/`range()` loop** — there's no way to avoid hitting this. Concretely, for:

```python
for x in range(3):
    print(x)
```

stepping the CSE Machine gives this `currentLine` trace (line 1 is the `for` line, line 2 is `print(x)`):

```
step 0: currentLine=1  (for x in range(3):)     ← correct — the only time line 1 ever appears
step 1: currentLine=1  (0)
step 2: currentLine=0  (3)                       ← BAD
step 3: currentLine=1  (1)
step 4: currentLine=0  (for)                     ← BAD — the loop condition re-check
step 5: currentLine=0  (assign)                  ← BAD
step 6: currentLine=0  (0)                       ← BAD
step 7: currentLine=0  (asgn x)                  ← BAD
step 8: currentLine=0  (print(x))                ← BAD
step 9-15: currentLine=2  (print(x) body)         ← correct
step 16-25: currentLine=0  (loop-continue machinery again) ← BAD, repeats every iteration
step 26+: currentLine=2  (print(x) body)          ← correct
```

**`currentLine` never becomes `1` again after step 0**, for the entire rest of execution — every single iteration's condition re-check and increment collapses it to `0` instead. Roughly half the steps in this trivial 3-iteration loop report the wrong line. The practical, user-visible effect: the `for` line's highlight lights up once on initial entry and then never again for any subsequent iteration, no matter how many times the loop runs.

## Suggested fix

Same pattern as #230's `serializeControlItem` fix: don't use `currentNode.startToken.line` when `currentNode.startToken.synthetic` is true. The real question is what *should* be shown instead while a synthetic node is executing — likely the line of the nearest enclosing real node (e.g. keep the previous non-synthetic `currentLine` rather than overwriting it with a synthetic one), since synthetic nodes don't correspond to any real line the user wrote.

## Repro

```python
for x in range(3):
    print(x)
```

Step through the CSE Machine and watch `currentLine` (or the editor's current-line highlight) — confirmed via `collectSnapshots` directly against py-slang on the #230 branch, and independently observed by @akshaydocs2007 in the actual frontend.

Akshay-2007-1:
## Summary

Closes #233.

`currentLine` in `collectSnapshots` (`src/conductor/plugins/PyCseMachinePlugin.ts`) drives the CSE Machine visualizer's "current line" highlight. It read `context.runtime.nodes[0]?.startToken?.line` directly, with no guard for runtime-fabricated nodes — the implicit `range()` start/step bounds (`utils.ts`'s `evaluateForIterator`) and the per-iteration loop-variable increment (`utils.ts`'s `generateForIncrement`, plus the synthetic nodes built in `interpreter.ts`'s `FOR` instruction handler). Those tokens are always constructed with `line: 0`, since they don't correspond to any line the user actually wrote.

`generateForIncrement` runs on **every iteration of every `for`/`range()` loop**, so this wasn't a rare corner case. For:

```python
for x in range(3):
    print(x)
```

`currentLine` collapses to `0` for roughly half the steps of even this trivial 3-iteration loop, and — worse — the `for` line's reported line never recovers after the very first step: every subsequent condition re-check and increment overwrites it with `0` again. The user-visible effect is that the `for` line's highlight lights up once on initial entry and then never again for the rest of the loop's execution.

## Fix

Track the last real (`line > 0`) line seen across the snapshot stream, and keep reporting it whenever the current node's own line is `0`, instead of overwriting with the meaningless synthetic value. Real tokens are always 1-based (see `parser/token-bridge.ts`'s `mooToken.line ?? 0` — Moo produces 1-based lines), so `line === 0` is an unambiguous "not a real line" signal on its own; no dependency on any synthetic-token flag/metadata is needed.

```ts
const rawLine = currentNode?.startToken?.line;
if (rawLine) {
  lastKnownLine = rawLine;
}
// ...
currentLine: lastKnownLine,
```

## Test plan

- Added a regression test in `src/tests/PyCseMachinePlugin.test.ts`'s existing `collectSnapshots` describe block: steps a `for x in range(3): print(x)` snippet and asserts `currentLine` is never `0` across any snapshot, and that it does reach line 2 (the loop body) rather than getting stuck.
- `npx tsc --noEmit` clean.
- `npx eslint` clean on changed files.
- Full `npx jest`: all non-skipped suites pass, matching a clean `main` checkout (one suite is skipped in this environment on `main` too, unrelated to this change).

### DIFF
--- a/src/conductor/plugins/PyCseMachinePlugin.ts
+++ b/src/conductor/plugins/PyCseMachinePlugin.ts
@@ -411,6 +411,13 @@ export async function collectSnapshots(
   maxSnapshots: number = 1000,
 ): Promise<CseSnapshot[]> {
   const snapshots: CseSnapshot[] = [];
+  // Runtime-fabricated nodes (e.g. implicit range() bounds, the for-loop increment —
+  // see utils.ts's evaluateForIterator/generateForIncrement) carry tokens pinned to
+  // line 0, since they don't correspond to any line the user actually wrote. Real
+  // tokens are always 1-based (see parser/token-bridge.ts), so line 0 is an
+  // unambiguous "not a real line" signal. Keep showing the last real line instead of
+  // flickering to 0 while these synthetic nodes are being evaluated.
+  let lastKnownLine: number | undefined;
 
   const stream = generateCSEMachineStateStream(
     code,
@@ -452,14 +459,17 @@ export async function collectSnapshots(
     // machine's updateInspector, which reads context.runtime.nodes[0] for the blue
     // "current line" highlight. py-slang nodes carry a 1-based startToken.line.
     const currentNode = context.runtime.nodes[0] as { startToken?: { line?: number } } | undefined;
-    const currentLine: number | undefined = currentNode?.startToken?.line;
+    const rawLine = currentNode?.startToken?.line;
+    if (rawLine) {
+      lastKnownLine = rawLine;
+    }
 
     snapshots.push({
       stepIndex: steps - 1,
       control: controlItems,
       stash: stashItems,
       environments,
-      currentLine,
+      currentLine: lastKnownLine,
     });
   }
 
--- a/src/engines/cse/interpreter.ts
+++ b/src/engines/cse/interpreter.ts
@@ -1121,14 +1121,17 @@ const cmdEvaluators: CmdEvaluators = {
         return;
       }
       control.push(instr);
+      // Line is the real for-statement's line — this per-iteration bookkeeping is
+      // logically part of evaluating that statement, not a separate, lineless step.
+      const forLine = node.startToken.line;
       const generateBigIntLiteral = (value: bigint): ExprNS.BigIntLiteral => {
-        const token = new Token(TokenType.BIGINT, value.toString(), 0, 0, -1);
+        const token = new Token(TokenType.BIGINT, value.toString(), forLine, 0, -1);
         token.synthetic = true;
         return new ExprNS.BigIntLiteral(token, token, value.toString());
       };
       const v1Lit = generateBigIntLiteral(start.value);
       const v3Lit = generateBigIntLiteral(step.value);
-      const plusToken = new Token(TokenType.PLUS, "+", 0, 0, -1);
+      const plusToken = new Token(TokenType.PLUS, "+", forLine, 0, -1);
       plusToken.synthetic = true;
       const nextStartExpr = new ExprNS.Binary(plusToken, plusToken, v1Lit, plusToken, v3Lit);
       Object.assign(nextStartExpr, { syntheticLabel: `${start.value}+${step.value}` });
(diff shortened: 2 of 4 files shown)


Akshay-2007-1:
@Shrey5132 opened this as a fix for #233. Tracks the last real (\`line > 0\`) line and reuses it whenever the current node's line is \`0\`, per your original suggested direction. Kept it as its own PR/branch off \`main\` rather than stacking on #230, since the fix doesn't need the \`synthetic\` flag from that PR at all — \`line === 0\` alone is an unambiguous signal, since real tokens are always 1-based.

Akshay-2007-1:
Pushed a refinement (b4c26b6): the initial fix kept showing the last real line while a synthetic node was executing, which eliminated the \`0\` bug but left \`currentLine\` stuck on the loop body's line forever after the first iteration, since the loop's condition-recheck/increment machinery is entirely synthetic from there on.

Instead, \`evaluateForIterator\`'s implicit \`range()\` bounds, \`generateForIncrement\`, and the \`FOR\` instruction handler's per-iteration nodes now carry the real \`for\`-statement's line (they're logically part of evaluating it) instead of \`0\`/nothing. \`currentLine\` now correctly alternates between the \`for\` line and the body line across iterations instead of sticking to one. Verified via trace on \`for x in range(3): print(x)\`, and independently confirmed in the actual frontend.

Shrey5132 (approved):

PR #235, opened 2026-07-08
CSE machine: allocate local variables as unassigned at CALL time

### LINKED ISSUE #223: CSE machine: Local variables must be allocated by CALL
Consider:
```
z = "global"
def f():
    g = lambda: z
    return g()
    z = "local"
print(f())
```
This program prints `global` in py-slang but throws a (somewhat cryptic) runtime error in CPython:
```
NameError: cannot access free variable 'z' where it is not associated with a value in enclosing scope
```
The name `z` gets accessed in the temporal dead zone: The name `z` must be in the environment of the call as soon as `f` is CALLed.  Upon function entry, the local variables need to be **unassigned**, i.e. not associated with a value. Any attempt to access an unassigned variable should throw this kind of runtime error.

In the current CSE machine, the allocation of the local `z` gets delayed until `z = "local"` which of course never happens in this example.

Image

In Python, only the global environment supports dynamic allocation.

Akshay-2007-1:
## Summary

Fixes #223.

Local variables were being bound lazily whenever their assignment statement happened to execute, rather than being allocated up front when the function is called. This meant a read that raced ahead of its own binding (e.g. inside a closure created before the assignment runs) would fall straight through the empty frame and resolve to an unrelated binding of the same name further up the environment chain, such as a global, instead of raising an error like CPython does.

Repro from the issue:

```python
z = "global"
def f():
    g = lambda: z
    return g()
    z = "local"
print(f())
```

CPython raises `NameError: cannot access free variable 'z' where it is not associated with a value in enclosing scope`. py-slang was printing `global`.

## Changes

- `environment.ts`: added an `UNASSIGNED` sentinel. `createEnvironment` now preallocates every local of a function frame (assignment/def/for targets, minus params already bound) as `UNASSIGNED` at CALL time, instead of leaving the slot missing until the first assignment executes.
- `utils.ts`: `pyGetVariable` and `pyGetNonlocalVariable` now check for the `UNASSIGNED` sentinel and raise `UnboundLocalError` when the owning frame is the current frame, or `FreeVariableUnboundError` when it's an enclosing frame, matching CPython's distinction. `pyGetGlobalVariable` is unaffected since the global frame is never preallocated (only the global scope supports dynamic allocation, per the issue).
- `PyCseMachinePlugin.ts`: the CSE visualizer's `serializeValue`/`serializeEnvChain` now handle the `UNASSIGNED` sentinel so an unassigned local renders sensibly in the frame display instead of crashing or showing garbage.
- `nonlocal.test.ts`: added the issue's repro as a test case, plus the mirror case where the assignment does run before the closure read, to confirm the local resolves normally in that path.

## Test plan

- [x] `yarn jest src/tests/nonlocal.test.ts` — all passing, including the two new cases for this issue
- [x] `yarn jest` (full suite) — passing aside from one pre-existing flaky timeout in `pvml-runner.test.ts` (subprocess test, unrelated to this change, confirmed to pass in isolation both with and without this diff)
- [x] `yarn lint` — no errors

### DIFF
--- a/src/conductor/plugins/PyCseMachinePlugin.ts
+++ b/src/conductor/plugins/PyCseMachinePlugin.ts
@@ -7,7 +7,7 @@ import type {
 import { Closure } from "../../engines/cse/closure";
 import { Context } from "../../engines/cse/context";
 import { Control } from "../../engines/cse/control";
-import { Environment } from "../../engines/cse/environment";
+import { Environment, UNASSIGNED } from "../../engines/cse/environment";
 import { generateCSEMachineStateStream } from "../../engines/cse/interpreter";
 import { Stash, Value } from "../../engines/cse/stash";
 import { InstrType, operatorTranslator, typeTranslator } from "../../engines/cse/types";
@@ -83,7 +83,10 @@ function formatValue(v: Value): string {
   }
 }
 
-function serializeValue(v: Value, envId = ""): SerializedValue {
+function serializeValue(v: Value | typeof UNASSIGNED, envId = ""): SerializedValue {
+  // A local that createEnvironment preallocated at CALL time but that hasn't been
+  // assigned yet — mirrors js-slang's uninitialized-`const`/`let` placeholder rendering.
+  if (v === UNASSIGNED) return { displayValue: "", label: "unassigned" };
   if (v === undefined || v === null) return { displayValue: "None", label: "NoneType" };
   const base = { displayValue: formatValue(v), label: typeTranslator(v.type) };
   if (v.type === "closure") {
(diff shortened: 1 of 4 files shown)


Akshay-2007-1:
On the coverage drop: it's -0.007%, which given the size of this codebase is just the handful of new branches in this diff (the UNASSIGNED-sentinel checks, the else split from the Gemini comment below) not being hit by every possible path, not a real regression. Not worth padding with tests just to nudge the percentage back up, the two new nonlocal.test.ts cases already cover the actual bug from #223 directly. Leaving it as is.

Also addressed the gemini-code-assist comment on pyGetVariable: added the explicit else. handleRuntimeError always throws (return type never) so it wasn't a functional bug, but the control flow reads cleaner now.

martin-henz (approved):
Looks good!

Issue #232, opened 2026-07-08
Specs: Revamping equality and identity


martin-henz:
The new, greatly simplified, specs: 

Python §1,2:

<img width="650" height="79" alt="Screenshot 2026-07-08 at 10 00 26 AM" src="https://github.com/user-attachments/assets/b9f08cfa-5dfc-4be1-9429-09926a1ef179" />

Python §3,4:

<img width="650" height="33" alt="Screenshot 2026-07-08 at 10 01 20 AM" src="https://github.com/user-attachments/assets/8de3117b-ff43-4f6d-a190-ddae111cbd64" />

<img width="853" height="91" alt="Screenshot 2026-07-08 at 10 02 34 AM" src="https://github.com/user-attachments/assets/e1b54127-c2ae-4e03-871b-bc724678ce53" />

PR #247, opened 2026-07-09
Remove equal now that == is structurally equal

### LINKED ISSUE #215: Remove equal function from Python x
Structural equality of == serves the same purpose as equal in SICP and SICP JS.

martin-henz:
Fixes #215.

Branches from and blocked by https://github.com/source-academy/py-slang/pull/231

## Summary
`==` now serves the same purpose `equal` did (as in SICP and SICP JS), once the equality revamp on this branch (#231, for #232) gave the CSE machine, PVML compiler, and substitution stepper structural comparisons on pairs/lists/None. This PR removes the now-redundant `equal` builtin everywhere it existed:

- `src/stdlib/linked-list.prelude.ts` (§2) and `src/stdlib/list.prelude.ts` (§3/4) — the two prelude definitions (PVML/Pynter share these preludes, so this covers that engine too).
- `src/conductor/stepper/lists.ts` — the substitution stepper's own independent `equal` reimplementation, plus its `and`/`not`/`isPairOf` helpers that only `equal` used (now dead).
- `docs/lib/linked_list.py` / `linked_list.js` — the JSDoc-style doc source stubs (not touching generated HTML under `docs/python/`, which is a build artifact).

**Base branch note**: this targets `equality` (#231), not `main`, because it's a real prerequisite — before that branch's stepper fix, the stepper's own `==` only supported `(numeric, numeric)` and `(string, string)` operands and could not stand in for `equal`'s structural comparisons on pairs/lists/None. I verified this directly before and after checking out `equality`:
```
# before (main):      llist(1,2,3) == llist(1,2,3)  →  stuck (never reduces)
#                      None == None                  →  TypeError
# after (equality):   llist(1,2,3) == llist(1,2,3)  →  True
#                      None == None                  →  True
```

### Test changes
`equal(a, b)` was also used as a general deep-equality assertion helper throughout ~100 test snippets in `linked-list.test.ts`, `list.test.ts`, `stream.test.ts`, `stdlib.test.ts`, and `getSteps.test.ts` — none of that was testing `equal` itself, so those are mechanically rewritten to `a == b`.

Two dedicated `equal`-behavior blocks are dropped rather than converted:
- §2's "equal semantics" tested `equal`'s bool-as-int coercion (e.g. `equal(True, 1)` → `True`), which `==` deliberately does **not** offer at §2 — bool is excluded from `==` there by design (see `MIDDLE_2` in `operator-spec.ts`, and the sweep in `operator-conformance.test.ts`). This was already an intentional inconsistency between `equal` and `==` at §2; removing `equal` resolves it in `==`'s favor, matching the pedagogical design. Its one non-bool case (`equal(None, False)` → `False`) is already covered by the §2 equality conformance sweep (bool × any → error).
- `stdlib.test.ts`'s §3/4 "equal on lists" block is dropped as fully redundant with the adjacent `==` parity coverage already there.

## Test plan
- [x] `npx jest` (full suite) — 2513/2513 passing (non-skipped), including `src/conductor/stepper` (179/179)
- [x] `npx tsc --noEmit` — no type errors
- [x] `grep -rn '\bequal(' src docs/lib docs/specs` — no remaining references anywhere in source, tests, or doc stubs


martin-henz:
Good catch, fixed in 13b9a39 — `or` was the last helper still hanging around after `equal` was removed (`and`/`not` were already dropped for the same reason in this PR). Confirmed zero remaining call sites before removing, full suite still green (2530 tests).

PR #250, opened 2026-07-09
Add print_llist to the stepper and PVML, matching the CSE machine

### LINKED ISSUE #249: Stepper: `print_llist` missing in Python 2
```
print_llist(pair(2,None))
```
currently gives
```
NameNotFoundError at line 1
                   
print_llist(pair(2,None))
 ^^^^^^^^^^^ This name is not found in the current or enclosing environment(s).
```
in Python 2.

martin-henz:
## Summary

`print_llist` (box-and-pointer text — `llist(a, b, c)` for a proper list, `[head, tail]` otherwise) was only implemented natively for the CSE machine (`src/stdlib/linked-list.ts`). Checked directly rather than assuming:

- **CSE**: already correct.
- **Stepper**: `print_llist` wasn't recognized as a builtin at all — a call just sat unreduced ("Evaluation stuck").
- **PVML**: failed at *compile time* with `Primitive function print_llist not implemented` — `pair`/`head`/`tail`/`llist`/`is_pair`/`is_llist` all work, `print_llist` specifically was missing from the primitive registry.

This PR adds it to both, as an independent duplicate of the algorithm in each engine (discussed and deliberately chosen over a shared cross-engine core, given the three engines use three different concrete value representations — `Value`/`StepNode`/`PVMLBoxType` — with no existing shared interface).

## js-slang#1124

While working on this I recalled [source-academy/js-slang#1124](https://github.com/source-academy/js-slang/issues/1124): `display_list` (js-slang's equivalent) prints the wrong notation for a value reachable via two different paths in the same structure (a shared sub-list gets misclassified). I checked whether we have the same bug class, in all three engines:

```python
x1 = llist(2, 3)
x2 = llist(x1, pair(1, x1))
print_llist(x2)
# correct: llist(llist(2, 3), llist(1, 2, 3))
```

We don't. The root cause in js-slang appears to be identity-keyed memoization (`Map<Value, StringDag>` in `stringify.ts`) that can reuse a stale classification for a shared node. Every implementation here — the existing CSE one and the two new ones — is plain, uncached recursion with no memoization keyed by value/object identity, so the same node is always re-derived fresh from its own structure regardless of which parent reached it first. Added this exact repro as a regression test to all three engines' suites.

## PVML native-Pynter caveat

`PRIMITIVE_FUNCTIONS`'s indices aren't arbitrary — they must match `sivmfn_primitives[]`'s position in the **native Pynter binary** (a separate repo, `source-academy/pynter`), since compiled bytecode bakes the index in directly. `print_llist` is registered at index 97, following the exact precedent already set by `is_integer`/`is_float`/`is_complex`/`_gen_list`/`arity` (92–96): "added as new native primitives appended to the end of Pynter's dispatch table." This makes it fully correct and tested for py-slang's own local TS-based `PVMLInterpreter` (what `pvml-interpreter.test.ts` exercises), but it will **not** work through the actual native Pynter binary until that separate repo adds a matching entry — out of scope for this PR.

## Test plan

- [x] `npx jest` (full suite) — 2535/2535 passing (non-skipped), including stepper (181/181) and PVML (97/97)
- [x] `npx tsc --noEmit` — no type errors
- [x] Verified directly: `print_llist` output matches across CSE/stepper/PVML for proper lists, `None`, improper pairs, nested proper-list elements inside improper pairs, string elements, arity errors, and the js-slang#1124 shared-substructure repro

fixes https://github.com/source-academy/py-slang/issues/249


martin-henz:
Fixed in 846df8d, same fix applied to both files as suggested. Made \`isProperLlist\` iterative and threaded an \`isKnownImproper\` flag through \`printLlistText\` so a tail position already known to continue an improper chain skips the redundant re-check.

Verified the actual impact before/after: a 3000-element improper chain went from ~18s to ~2ms; a 50,000-element proper list (previously untested at that scale) now completes in single-digit ms with no stack error. Added regression tests for both engines at those sizes, built directly via each engine's AST/value constructors so the measurement isn't confounded by the stepper's own per-statement execution overhead or PVML compilation limits.

One caveat carried over from your own analysis: a genuinely *improper* chain is still bounded by the bracket notation's own inherent O(N) stack depth (`[1, [2, [3, ...]]]` is N levels of nesting no matter how it's built) — that's accepted, matches the CSE machine's equivalent recursion-depth ceiling, and is unrelated to the O(N²)/redundant-check bug this fixes.

Issue #259, opened 2026-07-12
Native Pynter: 62 known gaps + 2 newly-found categorical limitations, found via generateNativePynterTestCases


martin-henz:
Companion to #258 (the same investigation for the PVML-in-browser pathway). Running `yarn pynter:report` against a freshly-built native `runner` binary on `main` (not the `pvml-browser-revival` branch) surfaced two new *categorical* gaps and 62 genuine one-off bugs. Both are now handled: the categorical ones via new `NATIVE_PYNTER_SKIP_REASONS` predicates, and the one-off ones via a new `knownGaps` parameter on `generateNativePynterTestCases` (mirroring `generatePvmlInBrowserTestCases`' identical mechanism from #258), plus a hand-rolled equivalent in `operator-conformance-pynter.test.ts`.

## Two new categorical gaps (added to `NATIVE_PYNTER_SKIP_REASONS`)

- **For-loops are entirely unsupported.** `PVMLCompiler.visitForStmt` compiles to `NEWITER`, whose opcode index is above `PYNTER_OPCODE_MAX` — so `assemble()` unconditionally rejects *any* program containing a `for` loop when targeting Pynter, regardless of the loop's contents. This alone explained 18 of the original 90 failures (nearly all of `loops.test.ts`, plus a few `nonlocal.test.ts`/`global-keyword.test.ts` cases whose actual scoping behavior was never reached because the program didn't even compile).
- **`apply_in_underlying_python()` is a Python §4-only feature.** It lives in the `parser` stdlib group, which `VARIANT_GROUPS` (runner.ts) only wires up at §4 — Pynter's target is §3 specifically (see pynter/README.md), so this is out of scope the same way `parse()`/`tokenize()` already are, not a bug. Confirmed at runtime too: every call throws `NameNotFoundError` against the binary regardless of arguments, since there's no native-Pynter primitive for it either. Explained the remaining 10 of the original 90.

That leaves 62 genuine, uncategorized bugs/gaps, skip-gated as exact-match `knownGaps`:

## `round()`

- The two-argument form (`round(x, ndigits)`) throws instead of rounding to `ndigits` places.
- `round(2.5)` returns `3`, `round(-2.5)` returns `-3` — Python's banker's rounding (round-half-to-even) isn't implemented; Pynter rounds half away from zero instead.

## `is` / `is not` — real VM-level identity bugs

[...]

Issue #258, opened 2026-07-12
PVML-in-browser: 103 known gaps vs. CSE reference, found by generatePvmlInBrowserTestCases


martin-henz:
Filed alongside PR #236, which adds `generatePvmlInBrowserTestCases` (src/tests/utils.ts) — a harness that reruns the same `TestCases` tables already used by `generateTestCases()`/`generateNativePynterTestCases()` against `PVMLInterpreter` (the pure-TypeScript PVML-in-browser VM), pinning it against the CSE machine as the reference implementation. It surfaced 103 real behavioral gaps, now skip-gated in the test suite (search for this issue number in the skip reasons) rather than left red, so CI stays green while the gaps stay tracked and visible as labeled skips instead of being silently hidden.

## Highest priority: prelude-closure-linking crash

**`Function at index N not found`** — thrown by `PVMLInterpreter.createClosure` (pvml-interpreter.ts:1232) — hit by nearly every `linked-list.test.ts` and `stream.test.ts` case that calls a *recursive* or *sibling-referencing* prelude-defined function (e.g. `append`, `remove`, `llist_to_string` from linked-list.prelude.ts; `stream_map`, `eval_stream`, `sieve` from stream.prelude.ts). Minimal repro:

```python
def llist_to_string(x): ...  # prelude-defined, recurses into itself
print(llist_to_string(llist(1, 2)))
```
→ `Function at index 10 not found`

This looks like a real bug in how `generatePvmlInBrowserTestCases`'s (and generally, any two-stage prelude-then-script) compilation model links a prelude-defined closure's function-table index across the prelude/script compile boundary — accounts for the large majority of the 103 known gaps (roughly 45 of them). Worth investigating first; fixing it would likely clear most of the linked-list/stream failures at once.

## Missing/incomplete builtins

- `len()`, `max()`, `min()` don't support strings at all (`len('abc')`, `max('a', 'c', 'b')`) — only arrays/numeric args work.
- String subscripting (`s[0]`) isn't implemented — "Cannot index non-array value".
- `round(x, ndigits)`'s two-argument form isn't supported — "round() takes exactly 1 argument".
- `complex(x)`'s single-argument coercion form (from int/float/bool/another complex/a string like `complex("1+2j")`) isn't supported — only the two-arg `complex(real, imag)` form works.
- `__program__` isn't implemented.
- `arity()` throws "not implemented" for a few primitives: `math_ulp`, `math_nextafter`, `input`.
[...]

PR #260, opened 2026-07-12
Native Pynter: feature parity with py-slang (best effort) + testing + complex numbers

martin-henz:
[blocked by https://github.com/source-academy/py-slang/pull/236]

## Summary

Companion to #258 (the same treatment for the PVML-in-browser pathway). Running `yarn pynter:report` against a freshly-built native `runner` binary surfaced two new *categorical* gaps and 62 genuine one-off bugs, none of which were previously tracked or gated — the native-Pynter suite is opt-in (`PYNTER_RUNNER_PATH`) and never runs in CI, so these failures had no visibility and no mechanism to catch regressions. Full categorized writeup in #259.

- Two new `NATIVE_PYNTER_SKIP_REASONS` predicates (matching the existing complex-number/`parse()`/`tokenize()` pattern):
  - **for-loops**: compile to `NEWITER`, whose opcode is above `PYNTER_OPCODE_MAX` — `assemble()` unconditionally rejects any program with a `for` loop when targeting Pynter. Explained 18 of the original 90 failures.
  - **`apply_in_underlying_python()`**: lives in the `parser` stdlib group, which is Python §4-only — Pynter targets §3, so this is out of scope the same way `parse()`/`tokenize()` already are. Explained the remaining 10.
- The remaining 62 are real bugs/gaps (round()'s two-arg form and banker's rounding, real `is`/`is not` identity bugs, NaN identity/equality — a structural limitation shared with #258's PVML-in-browser findings — missing arity validators, `arity()` miscounting, string-indexing breaking on multi-codepoint characters, wrong return values from `set_head`/`set_tail`/`print_llist`/`llist_to_string`, real `global`-keyword scoping bugs, and value bugs in `//`/`%`/`**`), skip-gated via a new `knownGaps` parameter on `generateNativePynterTestCases` — an exact-match list, not a predicate, since these are uncategorized one-offs with no single rule that could describe them without risking silently swallowing a future, different failure in the same table. Mirrors `generatePvmlInBrowserTestCases`' identical `knownGaps` parameter from #258.
- `operator-conformance-pynter.test.ts` doesn't use `generateNativePynterTestCases` (it's a hand-rolled sweep) — gets an equivalent `KNOWN_GAPS` Set + `testOrSkip()` helper for its 11 known-failing cases.

Every skip-gated case's exact code was taken from a fresh, JSON-reporter-verified run against the real binary — see #259 for the full list and categorization.

## Test plan

- [x] Cloned and built native Pynter's `runner` binary locally (CMake, `make -j8`) to get ground truth.
- [x] `PYNTER_RUNNER_PATH=<runner> yarn pynter:report` — **1157/1157 (100%)** across every suite, 182 labeled skips (120 categorical, 62 known-gap), zero unaccounted failures.
- [x] `npx tsc --noEmit`, `yarn lint`, `yarn format:ci` — clean.
- [x] `npx jest` (no `PYNTER_RUNNER_PATH` set, matching CI) — unaffected: 25/25 non-skipped suites pass, same pass count as before this PR, confirming this change is invisible to the normal (non-Pynter) test run.

## Update (2026-07-21): merged main + #303, fixed remaining gaps natively instead of skip-gating

Re-ran `yarn pynter:report` against a fresh native build after merging latest `main` and #303 (issue
#299's list-indexing/multiplication spec) into this branch — this surfaced two things the original
62-knownGaps sweep above didn't cover, since #303 was unmerged at the time:

- **~36 `stdlib` failures**: all the newer `math_*` stdlib additions (`comb`/`factorial`/`gcd`/
  `isqrt`/`lcm`/`perm`/`fabs`/`fma`/`fmod`/`remainder`/`copysign`/`isfinite`/`isinf`/`isnan`/`ldexp`/
  `exp2`/`gamma`/`lgamma`/`radians`/`degrees`/`erf`/`erfc`) had never been given a native Pynter
  primitive at all (`sivmfn_prim_unimpl` stubs) — plus the math functions Pynter *does* implement
  (`acos`/`acosh`/`asin`/`atanh`/`log`/`log1p`/`log2`/`log10`/`sqrt`) never checked CPython's domain
  restrictions (e.g. `sqrt(-1)` silently returned NaN instead of raising `ValueError`).
- **~15 `list` failures**: negative-index wraparound, out-of-range bounds checking, and bool/float
  index rejection were all missing from Pynter's list subscript opcodes, and `list * int`/`int * list`
  had no implementation at all — i.e. issue #299's spec, once its test content actually landed on this
  branch via the #303 merge.

Given Martin's direction ("we need to become more serious about pynter feature parity... change
pynter's C source accordingly, and not skip-gate"), both were fixed **natively in Pynter**
([source-academy/pynter#22](https://github.com/source-academy/pynter/pull/22)) rather than added to
`knownGaps` — no py-slang-side skip-gate entries were needed for either category as a result. The one
non-trivial architectural piece: Pynter's `STAG` opcode is shared between building list literals
(which relied on auto-growing an empty array) and real user subscript assignment (which must never
auto-grow, per #294/#299) — resolved by giving `NEWA` a size operand (this PR's
`opcodes.ts`/`pvml-assembler.ts`/`pvml-compiler.ts` changes) so literals pre-size their backing array
up front, matching what the browser-pathway `PVMLInterpreter` already did, and letting a single
strict bounds check be correct everywhere on the Pynter side.

Result: `yarn pynter:report` now shows **1395/1395 (100%)** of everything in scope (up from the
1157/1157 baseline above, which predates #303's test content), still zero unaccounted failures, and
full `yarn jest` (62 suites, 20611 tests) shows no regressions across every engine.

## Update (2026-07-22): rebuilt on a fresh `main`, complex numbers, rest parameters, and a real result-capture mechanism

Rebuilt this branch from scratch on top of a current `main` (resolving the merge conflict noted
above) rather than merging forward, since `main` had also independently gained an equivalent
opcode-allowlist mechanism and part of a `captureLastExpression`-style fix in the meantime. The net
result carries everything the branch already had (for-loop/`**`/`//` support, the `NEWA` size-operand
fix, the 62 previously-knownGaps entries — now re-verified against a real, currently-passing
comparison instead of relying on the old, silently-broken `resultType`/`resultValue` trailer) plus:

- **Complex numbers** ([source-academy/pynter#25](https://github.com/source-academy/pynter/pull/25)):
  full end-to-end support — `LGCC` now serializes for native Pynter (two inline float32s, matching
  Pynter's own `op_lgc_c`) instead of throwing, and Pynter's own VM gained a real `{real, imag}`
  heap type, complex-aware arithmetic/equality/identity, and `complex()`/`real()`/`imag()`/
  `is_complex()`.
- **A real `captureLastExpression` mechanism** (`pvml-runner.ts`): the previous version of this
  branch's own test harness turned out to be checking `resultType`/`resultValue`, which always
  report `"undefined"` for any bare expression on `main` — every native-Pynter test asserting a
  *value* (not just success/failure) was silently comparing `undefined` against `undefined`. This
  wraps a program's last bare expression in `print()` instead, so tests observe the real printed
  result. Re-running the full suite with this fix active surfaced several genuine bugs that were
  previously masked entirely (see below) — all fixed on the Pynter side, not skip-gated.
- **Rest parameters** (`def f(*args): ...`): removed the compile-time rejection for native Pynter;
  `pvm_function_t`'s unused padding byte is now `has_rest_param`, read by Pynter's `op_call`/
  `op_call_t` to bind fixed parameters and collect the remainder into a fresh array. Two previously
  skip-gated `arity()` cases (`arity((lambda *args: args))`, `arity((lambda x, *args: args))`) are
  real tests again. (Keyword-only parameters after a rest param, e.g. `def f(x, *args, z): ...`,
  remain out of scope — not spec'd for any Source Academy Python chapter at all, and not something
  CSE gives working semantics either; see py-slang#310.)
- **Real bugs found and fixed** once `captureLastExpression` made value-comparison actually work,
  all fixed natively in Pynter, not skip-gated: complex NaN self-equality/identity, `print_llist`/
  `set_head`/`set_tail` returning Sinter's `undefined` sentinel instead of Python's `None`,
  `is_array`/`is_list` not enforcing exact arity, `round()` (was aliased to Source's
  `math_round` — no banker's rounding, no `ndigits`), string indexing (`LDAG` only ever accepted an
  array operand), and a `range()` primitive-index bug (`30`, colliding with `list_to_string`, instead
  of its real slot `131` — this alone explained several for-loop test failures that looked like
  allow-list gaps but weren't).
- **`str()`/`repr()`**: added as real (deliberately minimal, leaf-values-only) native primitives,
  replacing the unimplemented `stringify`/`prompt` stubs — mainly to unblock
  `linked-list.prelude.ts`'s `llist_to_string`, which calls `repr()` internally.

Three follow-up gaps identified but intentionally left out of scope here, each with its own tracking
issue: str()/repr() still have no *dedicated* native-Pynter test coverage of their own (py-slang#308),
`operator-conformance-pynter.test.ts`'s comparison logic doesn't decode list-typed results at all, so
`list * int` isn't actually checked there despite Pynter supporting it (py-slang#309), and
`docs/specs/python_3_bnf.tex`'s `rest-names` grammar is defined but never wired into any production
(py-slang#310).

Verified: pynter's own `ctest` (34/34), full `yarn jest` (62 suites, 20899 tests) with
`PYNTER_RUNNER_PATH` pointed at a fresh Pynter build — no regressions across any engine.

### DIFF
--- a/src/engines/pvml/PVMLIRBuilder.ts
+++ b/src/engines/pvml/PVMLIRBuilder.ts
@@ -320,7 +320,7 @@ const STACK_EFFECTS = new Int16Array(OPCODE_MAX + 1);
   STACK_EFFECTS[OpCodes.STPB] = -1;
 
   // Array operations
-  STACK_EFFECTS[OpCodes.NEWA] = 1; // Takes no operand, produces an empty array (see visitListExpr)
+  STACK_EFFECTS[OpCodes.NEWA] = 1; // Operand is the array's pre-sized element count (see visitListExpr)
   STACK_EFFECTS[OpCodes.LDAG] = -1;
   STACK_EFFECTS[OpCodes.LDAB] = -1;
   STACK_EFFECTS[OpCodes.LDAF] = -1;
--- a/src/engines/pvml/builtins.ts
+++ b/src/engines/pvml/builtins.ts
@@ -47,24 +47,24 @@ import {
  * range() called outside a for-loop, and only for this file's own TS-side
  * PVMLInterpreter — native Pynter has no primitive for it at all.
  *
- * `str`/`repr` point at two of native Pynter's unimplemented stubs (index 90
- * `stringify`, index 91 `prompt`; both `sivmfn_prim_unimpl`) rather than
- * being left out entirely: SICPy preludes (e.g. linked-list.prelude.ts's
- * `llist_to_string`) reference `repr()`, and PVMLCompiler compiles a whole
- * file eagerly — leaving the name out would block every function in the
- * prelude from compiling, not just the one that calls it. Pointing them at
- * unimplemented-but-real stub slots lets compilation succeed for both
- * targets; a `targetsPynter`-compiled program calling either one still
- * faults cleanly if it ever reaches real Pynter, exactly as before. They
- * need genuinely different indices (not both 90, as before this file's own
- * TS-side str()/repr() implementation existed) because this file's
- * `executePrimitive` dispatches on primitive index alone, and str()/repr()
- * differ in one respect a shared index can't distinguish: a *bare* string
- * argument is quoted by repr() but not str() (e.g. `str("a")` -> `a`,
- * `repr("a")` -> `'a'`) — see cse-interop.ts/src/stdlib/utils.ts's
- * toPythonString. `prompt` (index 91, i.e. Python's `input()`) has no
- * PRIMITIVE_FUNCTIONS entry of its own, so borrowing its stub slot for
- * `repr` doesn't take anything away from a real feature.
+ * `str`/`repr` point at what were originally two of native Pynter's
+ * unimplemented stubs (index 90 `stringify`, index 91 `prompt`) — both now
+ * have a real, deliberately minimal implementation there (leaf values only:
+ * None/bool/int/float/complex/string/function fallback — no recursion into
+ * lists at the primitive level, though a prelude that recurses itself, e.g.
+ * linked-list.prelude.ts's `llist_to_string`, gets that for free via its own
+ * recursion + string concatenation). Neither str()/repr() nor this fact is
+ * covered by any native-Pynter test today — see py-slang#308 for the
+ * tracking issue on closing that gap. They need genuinely different indices
+ * (not both 90, as before this file's own TS-side str()/repr()
+ * implementation existed) because this file's `executePrimitive` dispatches
+ * on primitive index alone, and str()/repr() differ in one respect a shared
+ * index can't distinguish: a *bare* string argument is quoted by repr() but
+ * not str() (e.g. `str("a")` -> `a`, `repr("a")` -> `'a'`) — see
+ * cse-interop.ts/src/stdlib/utils.ts's toPythonString. `prompt` (index 91,
+ * i.e. Python's `input()`) has no PRIMITIVE_FUNCTIONS entry of its own, so
+ * borrowing its slot for `repr` doesn't take anything away from a real
+ * feature.
  *
  * `is_integer`/`is_float`/`is_complex` have no Source-standard-library
  * equivalent (Source's is_number() doesn't distinguish int/float, and Source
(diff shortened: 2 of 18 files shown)


martin-henz:
@coderabbitai review

PR #267, opened 2026-07-14
pynter-wasm: real npm dependency, opcode fix, and test coverage

martin-henz:
## Summary

Four commits, each doing one thing, building toward closing pynter-wasm's complete lack of test coverage:

1. **`85b9d35`** — `pynter-wasm.ts`: `null` (not `undefined`) for `cwrap`'s void return type, matching source-academy/pynter#18's Gemini-suggested fix — applying it literally surfaced that the *actual published package's* `--emit-tsd`-generated types have the same narrow `cwrap` signature as Emscripten's own upstream JSDoc, so this call site (unlike pynter's own README example) needs `undefined`. Also drops a now-redundant `as Uint8Array` cast the real types make unnecessary.
2. **`addc535`** — Consume `@sourceacademy/pynter-wasm` as a real dependency instead of vendored, stale, manually-copied files (which predated pynter's `siwasm_run` fix — the WASM build was completely non-functional before that landed).
3. **`ac405c3`** — Bump `PYNTER_OPCODE_MAX`: this branch's compiler doesn't emit anything Pynter (native or WASM, same VM source) doesn't already support — the previous ceiling predated floor-division/for-loop support landing in Pynter. Verified: `-7 // 3` now compiles and correctly returns `-3` (previously rejected outright).
4. **`208edc3`** — New `operator-conformance-pynter-wasm.test.ts`: closes the actual coverage gap. No test previously imported `pynter-wasm.ts`/`PyPvmlPynterEvaluator` at all, and the jest wasm mock's own comment ("Pynter WASM cannot run in Node") turned out to be stale. Unlike the native-Pynter suite, this one needs no `PYNTER_RUNNER_PATH` opt-in — pynter-wasm runs in-process, so it's a standing, always-on regression test now.

While building the test suite, found and filed **source-academy/py-slang#268**: `range()`/`for` loops fault at runtime (`invalid program`) via both native and WASM Pynter — isolated to `range()` specifically (`abs()`, `print()`, `math_sqrt()` all dispatch correctly), likely a `CALLP`-vs-`CALLV` or primitive-index mismatch between this compiler and Pynter's actual convention. Out of scope here since the operator-conformance sweep never touches loops/`range()` at all — tracked separately rather than blocking this PR.

## Test plan

- [x] `tsc --noEmit` clean, `eslint` clean.
- [x] New test file alone: 796 passed, 2 known-gapped (both investigated and explained in-file, not assumed), 0 failed.
- [x] Full suite: 3359 passed, 0 failed.
- [x] `EVALUATOR=PyPvmlPynterEvaluator rollup -c` succeeds; bundle's emitted `.wasm` asset byte-matches the installed package exactly (105071 bytes), confirming real `node_modules` resolution.


martin-henz:
@coderabbit review

PR #279, opened 2026-07-18
PVML-in-browser: support importing conductor modules (from X import y)

martin-henz:
Brings the PVML browser evaluator to parity with the CSE machine's module pathway: `from runes import heart` now loads the module through conductor's `ModuleLoaderRunnerPlugin`, converts its exports to PVML values, and binds them as globals before the chunk runs.

## Design

Follows the CSE machine's two-phase model (`evaluateImports` in `src/engines/cse/interpreter.ts`): **all async loading happens before execution**, so the parser/resolver/compiler need no changes at all — `FromImport` already compiles to nothing, and imported names already resolve to by-name `LDGG` loads against the persistent `globalEnv`, which `loadImports` now pre-seeds.

The one genuinely new mechanism is call-time async: `PVMLInterpreter.execute()` is fully synchronous, but conductor module functions are async. `executeAsync()` keeps per-instruction dispatch synchronous and suspends **only** on an actual module-function call — `dispatchCall` parks it in `pendingExtern`, the driver loop awaits it and pushes the result exactly where a primitive's would land (tail calls included, since `CALLT` is always followed by `RETG`). Programs without imports pay one field check per instruction; the sync `execute()` path is untouched for all existing callers.

## Changes

- **`src/conductor/PyDataHandlerEvaluator.ts`** (new): the conductor `IDataHandler` implementation extracted **verbatim** from `PyCseEvaluator` (which now extends it, behaviour unchanged) — `ModuleLoaderRunnerPlugin` requires an `IInterfacableEvaluator` (`IEvaluator & IDataHandler`), so `PyPvmlEvaluator` must carry the same interface.
- **`src/engines/pvml/modules.ts`** (new): `PVMLBoxType` ↔ `TypedValue` converters mirroring `cse/modules.ts`'s mapping decisions (NUMBER↔float, none↔EMPTY_LIST, PAIR↔2-element array, CLOSURE↔callable wrapper).
- **`src/engines/pvml/types.ts`**: two new box variants — `PVMLOpaque` (module-owned handles, e.g. runes) and `PVMLExtern` (imported module functions).
- **`src/engines/pvml/pvml-interpreter.ts`**: `executeAsync()` as above. Module→Python callbacks (a student function handed to a module's higher-order export) re-enter synchronously via the existing `invokeValue` hook; a nested extern call from inside such a callback is rejected with a clear error rather than deadlocking.
- **`src/conductor/PyPvmlEvaluator.ts`**: async prelude/chunk pipeline + `loadImports`. Chunks without imports never touch the module loader, and the plugin registers lazily on first import.
- **`src/tests/pvml-modules.test.ts`** (new): 11 tests — constants, aliases, module-function calls, higher-order callbacks, opaque round-trips, PAIR conversion, REPL persistence of imports, chapter-1 support, and both error paths (unknown module / missing export).

## Known limitations (documented in code)

- A *literal* 2-element Python list `[a, b]` passed to a module converts as a dotted pair — PVML has no construction-site tag to disambiguate (CSE grew `listLiteralValues` for this; can follow if it matters in practice).
- A Python callback invoked *by* a module cannot itself call another module function (nested async re-entrance) — clear runtime error instead.

## Testing

Full suite passes (10,245 tests), typecheck, eslint, and prettier clean.


Akshay-2007-1:
The limitation listed above where the result of a module call cant be passed into a module call again, doesnt exist anymore as I have fixed the double re-entry async issue allowing multiple nested module calls which are of utmost importance for sound and many other modules.

Akshay-2007-1 (approved):
LGTM!

#294 could be a followup!

PR #282, opened 2026-07-19
py2js: compile-to-JavaScript evaluator for Python §1-4

martin-henz:
## Summary

A fourth execution engine (alongside CSE, PVML and WASM): **py2js** compiles the py-slang AST to JavaScript source and runs it natively, in the style of js-slang's transpiler. Exec-style only (no last-expression display, matching Python semantics). Now covers the **full SICPy language, chapters 1-4**.

This PR grew via several stacked follow-ups merged into this branch (#287, #288, #290, #291, #296 — the last of which itself absorbed a chapter-4 follow-up, #297) plus a series of direct module-interop and representation fixes — this description now covers the whole engine surface for a single review pass. Full design writeup: **[`src/engines/py2js/README.md`](https://github.com/source-academy/py-slang/blob/py2js-engine/src/engines/py2js/README.md)** (added in this PR, kept up to date as the design changed) — this description summarizes; the README has the depth (exact mechanisms, file/line pointers, measured numbers).

### Design highlights

- **Unboxed value model** (`runtime.ts`): int→`bigint`, float→`number`, bool→`boolean`, str→`string`, None→`null`, complex→`PyComplexNumber`, list→`PyList` (a plain `PyValue[]`). A chapter-2 pair and a chapter-3+ list literal are the *same* `PyList` — one representation for both, matching the CSE machine exactly (which represents both as the same flat `{type:"list", value: Value[]}`, no separate pair type there either). "Is this a pair" is a structural question (length === 2), not a type-level one, on both engines. Operator/stdlib semantics mirror `evaluateBinaryExpression`/`evaluateUnaryExpression` in the CSE machine — same dispatch order, same per-chapter typing rules — so the engines agree by construction, not by a second hand-maintained copy.
- **Three compile modes**: sync (fastest, no async anywhere), dual (every user function gets two bodies over one closure environment — `__py.def2` — an async spine body and a sync body TS modules call back into at full speed), and REPL (one chunk of a persistent `Py2JsSession`, module-level bindings live in a runtime globals table instead of `let`s, for CSE-parity late-binding across chunks).
- **No JS TDZ vs. Python's dynamically-growing environments** — the trickiest piece of the whole engine. Every scope's names are hoisted as an uninitialized `let` before any code in that scope runs, with guarded `=== undefined` reads raising `UnboundLocalError`/`NameError` exactly where Python would, dynamically, regardless of textual position. `global`/`nonlocal` (chapter 3+) generalize this: a name a function declares `global`/`nonlocal` is excluded from its own hoisted locals, so the existing scope-chain walk falls through to the outer binding — including a whole-program pre-scan for a global a function introduces with no top-level assignment anywhere in the program. See the README's "No JS TDZ..." section for the full mechanism.
- **for-loops** desugar exactly per `docs/specs/python_loops.tex` (a hidden counter distinct from the user-visible target, advanced via a `for(;;)` update clause specifically so `continue` doesn't skip it and hang the loop).
- **Proper tail calls**, unbounded recursion depth (verified to 1M), via a trampoline + `__py.tail(f, args)` markers in return position.

### Pairs and lists are one representation, not two

Originally, chapter-2 pairs (`PyPair`, a dedicated cons-cell class) and chapter-3+ list literals (`PyValue[]`) were two distinct JS types — a deliberate choice at the time, reasoning that `Array.isArray` gave a free discriminant. In practice that split kept leaking out as real conformance gaps against the CSE reference, which has no such split: `pair(1, 2) == [1, 2]` was `False` in py2js but `True` on CSE (Gemini review finding, fixed with a bolted-on cross-type case at the time), and `pair(1, 2)[0]` threw `TypeError` in py2js but works on CSE (`listAccess`/`listAssign` only ever learned to accept `Array.isArray`, never `PyPair`).

Rather than patch each site individually, `PyPair` is now deleted entirely — a pair is just a 2-element `PyList`, matching CSE exactly, no separate type to keep in sync. Every site that used to branch on `instanceof PyPair` vs `Array.isArray` collapses into one branch: `pyEquals` (the cross-type case from the Gemini fix is now dead code, deleted — there's only one type again), `toDisplayValue`, `pyStr`, `truth`, `pyTypeName`, `stdlibBridge.ts`'s `toTagged`/`fromTagged`/`nativeSetPairSlot`/`nativeStream`/`walkArgList`, `moduleInterop.ts`'s `pythonToModule`/`moduleToPython`. `isProperList`/`printLlist` (the `print_llist` box-and-pointer renderer) were rewritten to walk 2-element-array chains instead of `PyPair` chains, checked against CSE's actual `_is_llist`/`_print_llist` boundary logic (`src/stdlib/linked-list.ts`) to get the edge cases right — a coincidentally-2-element literal list renders as bracket notation exactly like an improper pair does, and a genuine N-element (N≠2) list is never mistaken for a chain at any step.

Per explicit request: error messages now say **"pair" at chapters 1-2 and "list" at chapter 3+** for a list-shaped value, matching CSE's `friendlyTypeName(name, variant)` exactly (verified against a live CSE run: `"unsupported operand type(s) for +: integer and pair"` at chapter 2 vs. `"...and list"` at chapter 3, for the identical `1 + pair(1, 2)`). `pyTypeName` gained an explicit `sayPair` parameter threaded from whichever `restrict`/`universalEquality`-style chapter signal was already in scope at each of its 11 call sites — no new plumbing needed.

Two side-effect fixes found while doing this, not gone looking for separately: `toTagged`'s iterative tail-spine walk (there so a long pair/`llist()` chain doesn't overflow the stack converting into a CSE `Value`) only existed in one direction before — `fromTagged`'s list case (CSE → py2js) now gets the same protection, symmetrically. And `pythonToModule`'s "object" case only ever recognized `instanceof PyPair` for crossing a pair into a module — a genuine chapter-3 list literal fell through to a misleading "complex values are not supported" error; unifying the type check to `Array.isArray` fixes this as a direct consequence (now covered by a test), with an explicit length-check added so an N-element list errors clearly instead of silently truncating via `dh.pair_make`.

Zero changes needed to any other test file for this refactor — the whole existing py2js suite (5000+ cases) passed unmodified against the new representation.

### Stdlib

Runs the **real stdlib** (`misc`/`math`/`linked-list`/`list`/`pairmutator`/`stream`/`parser` — the same implementations every other engine uses), bridged via a native↔tagged value conversion layer (`stdlibBridge.ts`) rather than a reimplemented set, so stdlib semantics can't drift between engines. Most builtins need zero special-casing — the CSE tagged value union's list shape is always nested 2-element cons cells, so the same conversion that handles a pair also reconstructs an arbitrary-length parse tree, which is why `tokenize`/`parse` (chapter 4) needed no new bridging code at all. A handful of builtins (`print`/`input`/`arity`, `set_head`/`set_tail`, `stream()`, `apply_in_underlying_python`) are native instead, because the generic bridge's fresh-value round-trip is structurally wrong for them (would silently drop a mutation, or can't wrap a closure a CSE builtin fabricates on the spot, or needs a synchronous return where the generic path expects control-stack side effects) — see the README for the specifics on each.

### Module interop

`moduleInterop.ts` (+ the shared, engine-agnostic `GenericDataHandler` in `src/conductor/GenericDataHandler.ts`) converts between py2js's native values and conductor's module protocol. Closures crossing the module boundary are `AsyncGenerator` calls by conductor's own contract, not an engine choice — but conductor modules load via dynamic `import()` into the *same* JS realm as the evaluator (confirmed by reading `runner-module-loader`'s actual implementation — the live evaluator object is passed straight into the module plugin's constructor, which a real worker boundary couldn't do without serialization), so the mandatory `AsyncGenerator` shell around a module→Python closure call (`pyClosureFunc`) already does a single direct `rt.callSync` inside it, not a real re-entry or cross-thread round-trip.

**`GenericDataHandler.closure_call_sync`**: a synchronous fast path for a closure that provably never needs a real host round-trip (the sound-module wave-sampling case — scalar in, scalar out). Even the mandatory `AsyncGenerator` shell costs real, measurable time at high call counts from pure allocation/microtask-scheduling overhead, not real work — measured in an isolated microbenchmark at ~16-20x (plain async function) to ~18-23x (matching `pyClosureFunc`'s actual async-generator shape) the cost of a plain synchronous call, for the exact same trivial computation. `closure_call_sync` looks up the stored closure and, if it carries a `.sync` twin (which `pyClosureFunc` now sets, backed by restricted-but-genuinely-synchronous scalar-only value converters), calls it directly — no generator, no `Promise`. Falls back cleanly (`undefined`) for anything without one; a result that can't be represented synchronously throws rather than silently falling back, since by that point `rt.callSync` has already run real Python code with real side effects and re-running it would double-invoke.

Also fixed: module `PAIR` values round-trip through a 2-element `PyList` (and, per the representation-unification above, so does a genuine chapter-3+ list literal passed into a module — there's no longer a type distinction for this conversion to even trip over), and a module closure passed back into a module (e.g. a `Sound`'s wave function created by one module call and later sampled by another) is passed through by its original identifier instead of being incorrectly re-wrapped as a plain Python closure.

### Chapter coverage

| Chapter | Adds |
|---|---|
| 1 | arithmetic, conditionals, functions/recursion, lambdas, `misc`/`math` |
| 2 | pairs (a 2-element `PyList`), linked-list stdlib, `from X import y` module loading |
| 3 | reassignment, `while`/`for`/`break`/`continue`, native lists + subscript access/assignment, `global`/`nonlocal`, universal `==`/`!=`/`is`/`is not`, `list`/`pairmutator`/`stream` |
| 4 | `tokenize`/`parse`/`apply_in_underlying_python`, predeclared `__program__` (available at every chapter, matching the CSE reference — not gated to chapter 4 despite being documented there) |

## Test plan

- `yarn jest src/tests/operator-conformance-py2js.test.ts` — full operator × type × type cross product against a live CSE reference, chapters 1-3: ~2800 cases
- `yarn jest src/tests/stdlib-conformance-py2js.test.ts` — every bridged builtin/constant, same live-reference approach
- Directed suites per feature area: `py2js-loops`, `py2js-lists` (now including `pair(1,2)[0]`/`pair(1,2)[0]=99` matching CSE, `print_llist` boundary cases, and chapter-2-vs-3+ error wording, all verified against live CSE output), `py2js-global-nonlocal`, `py2js-identity`, `py2js-interpreter-support`, `py2js-unbound`, `py2js-dual-mode`, `py2js-from-import`, `py2js-module-interop` (now including the chapter-3-list-into-a-module round-trip), `py2js-host-functions`, `Py2JsEvaluator`, `GenericDataHandler` — several reuse the exact fixtures from the CSE/PVML/native-Pynter/CPython conformance suites (`loops.test.ts`, `list.test.ts`) rather than hand-rolled parallel cases
- `yarn jest` — full suite green (5000+ cases)
- rollup build: `dist/Py2JsEvaluator1.{js,cjs}` build cleanly

## Not in this PR / known limitations

- Error source locations are not attached yet.
- Negative list indexing (`xs[-1]`) is unsupported, matching a gap in the CSE reference itself (not fixed here in isolation, since that would make the two engines disagree).
- A bridged stdlib builtin error can render as `"[object Object]"` instead of a real message (#295) — pre-existing since chapter 2, tracked separately.
- `input()` is not implemented.
- `DataType.ARRAY` and complex numbers do not cross the module boundary.
- No chapter 5 — SICPy has four chapters.

https://claude.ai/code/session_01QhrHrHQHramQhK5sNAVp9r

### DIFF
--- a/experiments/py2js/README.md
+++ b/experiments/py2js/README.md
@@ -1,0 +1,175 @@
+# py2js — compile-to-JavaScript evaluator experiment
+
+A feasibility experiment for a fourth py-slang execution engine (alongside the
+CSE machine, PVML and WASM): compile SICPy **chapter 1** to JavaScript source
+and run it natively, in the style of js-slang's transpiler
+(`js-slang/src/transpiler/transpiler.ts`).
+
+```
+yarn tsx experiments/py2js/compare.ts       # differential conformance vs the CSE machine
+yarn tsx experiments/py2js/bench.ts         # benchmark vs CSE machine and PVML interpreter
+yarn tsx experiments/py2js/module-bench.ts  # sound-module scenario: dual compilation
+```
+
+Nothing here is wired into the build, tests, or exports — it is an experiment.
+
+## Design
+
+| piece | file | idea |
+|---|---|---|
+| runtime | `runtime.ts` | Unboxed JS values: int→`bigint`, float→`number`, bool→`boolean`, str→`string`, None→`null`, functions→JS closures with `pyName`/`pyArity` metadata. Python semantics live in `binop`/`unop`/`truth` helpers that mirror `src/engines/cse/operators.ts` at chapter 1 (§1 equality excludes bool/function; `and`/`or` need a bool left operand; if/ternary use Python truthiness, per the BRANCH instruction). Reuses `pythonMod`/`numericCompare` from `src/engines/cse/utils` and `toPythonFloat` from `src/stdlib/utils`, so numeric edge cases agree with the other engines by construction. |
+| compiler | `compiler.ts` | Direct AST→string codegen over `ExprNS`/`StmtNS` (~200 lines). User identifiers are mangled with `$` (impossible in Python identifiers), all runtime services hang off one `__py` parameter, so no free-identifier collision handling is needed — much simpler than js-slang's `getUniqueId` machinery. |
+| tail calls | both | Returns in tail position compile to `__py.tail(f, args)` markers; `__py.call` runs the trampoline. The transform recurses through ternaries, groupings and `and`/`or` right operands — the same scheme as js-slang's `isTail` transform, but done during codegen instead of by AST rewriting. |
+
+Generated code for `fact`:
+
+```js
+$fact = __py.def("fact", 1, ($n) => {
+  return (__py.truth(__py.binop("==", $n, 0n)) ? 1n
+          : __py.binop("*", $n, __py.call($fact, [__py.binop("-", $n, 1n)])));
+});
+__py.call($print, [__py.call($fact, [20n])]);
+```
+
+## Results
+
+**Conformance** (`compare.ts`): **59/59** chapter-1 programs agree with the CSE
+machine — value semantics (int/float coercion, floor-div/mod signs, `**`,
+Python float formatting incl. `-0.0`/`1e+30`), chained comparisons,
+short-circuiting, closures, higher-order functions, deep (100k) and mutual
+tail recursion, and 14 negative cases where both engines must raise
+(§1 equality restrictions, non-bool `and`, ZeroDivision, arity, …).
+
+**Performance** (`bench.ts`, Node 20, M-series):
+
+| workload | CSE machine | PVML interpreter | py2js |
+|---|---|---|---|
+| `fib(20)` | 258 ms (74×) | 128 ms (37×) | **3.5 ms** |
+| tail sum 100k | 1136 ms (205×) | 1156 ms (208×) | **5.6 ms** |
+| newton 100k (float) | 2154 ms (189×) | 2204 ms (194×) | **11.4 ms** |
+
+Against hand-written native JS, py2js is within 2.3–3.3× (`fib(27)`,
+tail-sum 3M). Parse+compile is ~0.5 ms for a small program — compilation cost
+is negligible next to CSE startup.
+
+## Limitations found
+
+- **Non-tail recursion depth**: real JS stack frames, overflow between 1k and
+  5k Python frames (each Python call costs several JS frames). Tail recursion
+  is unbounded thanks to the trampoline. Notably CPython itself raises
+  `RecursionError` at a default limit of 1000 — a real engine should count
+  frames and raise `RecursionError` deliberately rather than leak `RangeError`,
+  making this *more* faithful than the CSE machine's unbounded recursion, and
+  SICP-style iterative processes are tail-recursive anyway.
+- **No source locations in errors** yet. The js-slang answer: emit
+  line/col literals into the runtime-helper calls (and/or source maps).
+- **Builtins are a minimal native set** (`print`, `abs`, `str`, `max`, `min`,
+  a few `math_*`). The real stdlib (`src/stdlib/misc.ts`/`math.ts`) is written
+  against the CSE machine's *tagged* `Value` union with `(args, source,
+  command, context)` signatures. Building py2js in anger needs either a
+  wrap/unwrap bridge at builtin call boundaries or (better, long-term) a
+  representation-agnostic stdlib core.
+- Not covered: complex numbers, MultiLambda, and everything ≥ chapter 2
+  (lists, loops, reassignment — all of which map naturally onto JS arrays,
+  loops and mutation; loops would *reduce* the trampoline's importance).
+- The resolver/validators are not run (the CSE runner's `analyze()` should run
+  first in a real engine — same as runner.ts does).
+
+## The async/conductor-module question
+
+In production the program runs in a web worker under conductor, and imported
+modules talk to the frontend over **asynchronous** channels. Does the
+transpiled code have to be entirely async?
+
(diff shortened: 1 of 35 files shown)


martin-henz:
Addressed all four Gemini comments in 2e44be0:

- **`abs(complex)` → `Math.hypot`** — applied; it's also what CPython's `abs(complex)` uses, so this is more faithful, not just more robust. Note: the CSE machine's own `abs` (stdlib `misc.ts`) still computes `sqrt(re² + im²)`, so the two engines now differ on extreme components like `abs(complex(1e200, 0))` (py2js: `1e200`, CSE: `inf`). Aligning `misc.ts` to `hypot` is a candidate follow-up in a separate PR.
- **`math_floor` non-finite guard** — applied, raising `ValueError` for NaN and `OverflowError` for ±inf, matching CPython's `math.floor`.
- **Error-kind preservation in `Py2JsRunError`** (both call sites) — applied via `e.name` (which `Py2JsRuntimeError` sets to the Python error kind).

Operator conformance sweep (749 cases) and lint/format re-verified green.

PR #285, opened 2026-07-19
Fix #284: compute complex magnitudes with Math.hypot

martin-henz:
## Summary

`sqrt(re² + im²)` overflows to `inf` (or underflows to `0`) when a component's square leaves the float range, even though the true modulus is representable — `abs(complex(1e200, 0))` came out as `inf` instead of `1e+200`. `Math.hypot` scales internally and is what CPython's `abs(complex)` uses.

Fixes the three sites #284 lists:
- `src/stdlib/misc.ts` — the CSE machine's `abs`, complex case
- `src/engines/pvml/builtins.ts` — the PVML interpreter's `abs`, complex case
- `src/types/value-types.ts` — `PyComplexNumber.pow`'s magnitude, shared by every engine

This also restores cross-engine agreement with py2js (#282), whose `abs` already switched to hypot in Gemini review.

Fixes #284

## Test plan

- `src/tests/complex-magnitude.test.ts` (new): CSE and PVML `abs(complex(1e200, 0))` → `1e+200`; `(1e200+0j) ** 0.5` → `1e100` through `PyComplexNumber.pow`; an ordinary `abs(complex(3, 4))` → `5.0` regression check
- `yarn jest src/tests/stdlib.test.ts src/tests/pvml-interpreter.test.ts` green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FbExZBaxgeUL4oi1t7PG93

PR #287, opened 2026-07-19
py2js: bridge the real stdlib (misc + math) at the value boundary

martin-henz:
## Summary

Stacked on #282. First step of the chapter-1 depth work: py2js now runs the **actual stdlib** — the same `misc`/`math` builtin implementations the CSE machine uses — instead of its placeholder builtin set, so stdlib semantics cannot drift between engines.

- **`stdlibBridge.ts`** (new) — converts native py2js values (int=`bigint`, float=`number`, None=`null`, complex=`PyComplexNumber`, …) to the CSE tagged `Value` union at the builtin call boundary and back for the result. Functions cross one way, as inspectable stand-ins (a minimal `FunctionValue` for user functions, a stub `BuiltinValue` for py2js natives): `is_function`, `str()`/`repr()` rendering and error messages all agree with the CSE machine. Errors raised through `handleRuntimeError` propagate out under their class names; each bridged builtin gets a synthetic `Call` node carrying its own name, so callee-naming messages ("unsupported argument type for `math_sin`…") stay accurate. A `Promise` guard rejects any future async stdlib addition loudly.
- **`runtime.ts`** — the native builtin core shrinks to what the bridge cannot supply: `print` (stream-based/async in the stdlib; the native version pushes to the runtime's output with identical formatting), `input` (raises "not supported yet"), and `arity` (py2js functions aren't CSE closures; bridged builtins carry `pyMinArgs` so both engines report identical arities — argument-count validation itself stays with the stdlib's own `@Validate` wrappers, preserving its exact error messages).
- **`compiler.ts`** — lambdas are named `(anonymous)`, matching the CSE machine's rendering of nameless closures.
- **`index.ts`** — `prepare()` bridges `PY2JS_GROUPS[variant]` (misc + math at chapter 1); native core wins over same-named bridged entries; `extraBuiltins` (future module bindings) override anything.

### Table-driven parity: `stdlib-conformance-py2js.test.ts`

**896 cases, all passing**: every misc/math builtin and constant swept over the chapter-1 type universe — the empty call, one argument of each type, and int/float pairs — exercising every `@Validate` arity gate, every per-type dispatch arm, and both conversion directions; plus directed arity/str/composition cases. Each case compares against the CSE machine running the identical `print(...)`-wrapped program via `runCode`, computed fresh (same philosophy as the operator sweeps). Exclusions: `input` (no comparable outcome) and value-comparison for `random_random`/`time_time` (nondeterministic; success-vs-error only).

Note: with the bridge, py2js's `abs(complex)` follows `misc.ts` — parity with CSE is guaranteed either way, and the hypot fix for extreme components lands for all engines together via #285.

## Test plan

- `yarn jest src/tests/stdlib-conformance-py2js.test.ts` — 896/896
- `yarn jest src/tests/operator-conformance-py2js.test.ts` — 749/749 (unchanged, now with bridged builtin resolution)
- `yarn jest` — full suite green (44 suites, 15 777 tests)

### DIFF
--- a/src/engines/py2js/compiler.ts
+++ b/src/engines/py2js/compiler.ts
@@ -129,8 +129,10 @@ function emitExpr(e: ExprNS.Expr, a: boolean, dual: boolean): string {
     case "Lambda": {
       const l = e as ExprNS.Lambda;
       // A lambda body is in tail position: evaluating it *is* the return.
+      // "(anonymous)" matches the CSE machine's rendering of lambda values
+      // (toPythonString on a nameless closure), pinned by the stdlib sweep.
       return emitFunctionValue(
-        "<lambda>",
+        "(anonymous)",
         l.parameters.map(p => mangle(p.lexeme)),
         bodyAsync => emitTailPosition(l.body, bodyAsync, dual),
         dual,
--- a/src/engines/py2js/index.ts
+++ b/src/engines/py2js/index.ts
@@ -20,7 +20,7 @@ import misc from "../../stdlib/misc";
 import type { Group } from "../../stdlib/utils";
 import { makeValidatorsForChapter } from "../../validator";
 import { CompileMode, compileProgram, Py2JsCompileError } from "./compiler";
-import { Py2JsRuntime, Py2JsRuntimeError, PyValue } from "./runtime";
+import { annotateHostFunction, Py2JsRuntime, Py2JsRuntimeError, PyValue } from "./runtime";
 import { bridgeStdlibGroups } from "./stdlibBridge";
 
 /**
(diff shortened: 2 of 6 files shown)


martin-henz:
Both Gemini comments addressed in 868e4f3. They flagged the same underlying gap — a plain JS function entering py2js via `extraBuiltins` without `pyName`/`pyArity` metadata — so alongside the two suggested point fixes (`Function#length`/`name` fallbacks in `arity()` and `toTagged`), the root cause is fixed at the boundary: `prepare()` now runs every extra builtin through `annotateHostFunction`, which establishes the full `PyFunction` metadata invariant (name from the binding, `pyBuiltin` rendering, `pyMinArgs` from `Function#length` for `arity()` reporting).

One deliberate deviation from the suggestion: `pyArity` is set to `-1` (unenforced) rather than `Function#length`, because a rest-args host implementation reports `length` 0 and enforcing it would wrongly reject every call — `length` is used for `arity()` *reporting* only. New `py2js-host-functions.test.ts` pins the scenario (bare function callable, `arity()`, rest-args, rendering); all three py2js suites green (1 649 tests).

PR #300, opened 2026-07-20
Revert verbose builtin repr on CSE machine stash

### LINKED ISSUE #298: CSE machine: display of built-ins in stash
Image

it should say
`built-in`
to be consistent with the display of 
`closure`
if the function is user-defined.

The tooltip currently also says 
`'print'`
I think it should say 
`<built-in function print>`
to be consistent with what 
`print(print)` 
shows in the REPL.

Akshay-2007-1:
Closes #298.

PR #230 changed how builtins show up on the CSE machine stash, from the verbose `<built-in function print>` down to just the bare name `print`, to match how closures get shown there. Turns out that was actually a step backward, this should read `<built-in function print>` like it used to, so it's consistent with what `print(print)` shows in the real REPL.

Only reverting the display string part of that PR here (the `formatValue` case in `PyCseMachinePlugin.ts`). Left the `typeTranslator` label change alone since that one's now load bearing for the wrong-type error message overhaul that landed later, pulling that out would break those error messages instead of fixing anything.

Updated the two tests in `PyCseMachinePlugin.test.ts` that had baked in the old bare-name expectation.

## Test plan
- `npx tsc --noEmit` clean
- `npx eslint` clean
- `npx jest src/tests/PyCseMachinePlugin.test.ts` passes
- `npx prettier --write` run on both changed files

### DIFF
--- a/src/conductor/plugins/PyCseMachinePlugin.ts
+++ b/src/conductor/plugins/PyCseMachinePlugin.ts
@@ -72,11 +72,8 @@ function formatValue(v: Value): string {
       return `[${items.join(", ")}${suffix}]`;
     }
     case "builtin":
-      // Just the name, matching how "closure" above shows funcName rather than
-      // Python's full repr() — this is a compact stash chip, not repr() output.
-      // The "this is a builtin" fact is carried separately via `label`
-      // (see serializeValue / typeTranslator), same pattern as closures.
-      return v.name;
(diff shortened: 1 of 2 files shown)


Akshay-2007-1:
@codetabbitai, are these fixes complete or are we missing smth and smth else needs fixing as well?

martin-henz (approved):
LGTM

PR #306, opened 2026-07-21
Remove default PVML instruction limit

Akshay-2007-1:
## Summary
- PVML runs in a webworker, so the frontend can already terminate runaway execution at any time — the previous hardcoded 1,000,000-instruction cap was outdated (unlike py2js/the transpiler path, which has no worker boundary to fall back on).
- `maxInstructionLimit` now defaults to `Infinity`; callers can still opt into a cap via the existing `maxInstructions` constructor option (used by the "exceeding maxInstructions throws" test).

Per Martin Henz's feedback in Telegram (2026-07-20): "we should not have a 1,000,000 step limit. This is outdated: It's running in a webworker so we can stop the worker any time from the frontend."

## Test plan
- [x] `yarn jest src/tests/pvml-interpreter.test.ts` — 226 passed
- [x] `npx tsc --noEmit`
- [x] `npx eslint` / `npx prettier --write` clean

### DIFF
--- a/src/engines/pvml/pvml-interpreter.ts
+++ b/src/engines/pvml/pvml-interpreter.ts
@@ -143,7 +143,12 @@ export class PVMLInterpreter {
   private callDepth: number = 0;
 
   private instructionCount: number = 0;
-  private maxInstructionLimit: number = 1000000;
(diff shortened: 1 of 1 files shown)


martin-henz (approved):
infinity is a good step limit

PR #312, opened 2026-07-22
Add Pyodide evaluator: real CPython via WebAssembly, gated by py-slang validation

### LINKED ISSUE #11: Support Full Python using Pyodide

martin-henz:
## Summary
Adds `PyodideEvaluator1-4` + `PyodideEvaluatorFull`, running Python on a real CPython runtime (via [pyodide](https://pyodide.org)) rather than py-slang's own interpreter/compiler — full stdlib, real `import numpy`-style packages via micropip. Closes #11.

- `PyodideEvaluator1-4` gate every chunk through the same parser + Resolver (`analyze()`) every other engine uses, with that chapter's feature validators and stdlib groups — so "Python §N" still means the restricted SICPy subset even though CPython itself would happily run more. Only a chunk that passes validation actually runs, on pyodide, unmodified.
- `PyodideEvaluatorFull` skips the chapter gate entirely (unrestricted Python — "Python Full" in the language directory, not one of the four chapters). See companion fix in [language-directory#43](https://github.com/source-academy/language-directory/pull/43).
- Missing top-level imports are installed on demand via micropip; already-resolved roots are cached per evaluator instance.

## Relationship to #117
This replaces #117 (credit to @veehz for the original pyodide integration), which predates the evaluator-naming overhaul, the build-system rewrite (`scripts/build.ts`/`rollup.config.mjs`), and the conductor 0.3.0→0.7.0 bump — rebasing it in place wasn't practical. This PR ports the reusable core against the current architecture, **scoped to plain Pyodide only** — the PyTorch bridge from #117 (`bridge.py`, `loadTorch.ts`, `@sourceacademy/torch`) is deliberately left out; see the follow-up issue for picking that back up.

Two correctness issues in the original that this fixes along the way:
- A process-wide `helperLoaded` boolean in `importAnalyzer.ts` meant a second `PyodideEvaluator` instance's own pyodide runtime never actually got its import-analysis helper defined (each evaluator owns a separate pyodide runtime). Now tracked per-instance.
- `validateChunk()` called `analyze()` but never checked its returned errors array — chapter feature-gate violations were silently ignored. Now the first error is thrown and reported via `sendError`.

Also new (the original had neither): REPL-style cross-chunk name tracking, so `x = 5` in one chunk resolves correctly in a `print(x)` in the next — mirrors `PyPvmlEvaluatorBase`/`Py2JsSession`'s own `priorGlobals` handling, adapted for the fact that pyodide's actual runtime state (not a py-slang-side object) is the source of truth here.

## Testing note
pyodide's Node-side loader needs a dynamic `import()` that only works under `node --experimental-vm-modules` — which compiles the *entire* test tree as real ESM and removes `__dirname` (used throughout the rest of the suite, e.g. `src/tests/utils.ts`'s `CPYTHON_BATCH_RUNNER`). Confirmed empirically these are incompatible TypeScript module-target requirements, not just a runtime quirk (`import.meta` is a hard compile error under the non-flagged mode; `__dirname` is `undefined` under the flagged one). So `yarn test:pyodide` runs the two pyodide test files separately, under the flag, wired into CI as its own step — the default `yarn test`/`test-coverage` is untouched.

## Test plan
- [x] `yarn test:pyodide` — 11 tests passing (chapter gating, cross-chunk persistence, micropip auto-install, error propagation)
- [x] `yarn test` — full suite unaffected (3331 passing, same as main)
- [x] `yarn lint`, `tsc --noEmit` clean


martin-henz:
@coderabbitai review

Issue #321, opened 2026-07-22
WASM: set_timeout/clear_all_timeout support


martin-henz:
## Background

#311 added `set_timeout(f, t)` / `clear_all_timeout()` to the MISC stdlib group (needed for the tone matrix quest, Q5B), implemented for py2js only so far — see that PR. Calling `set_timeout` under the WASM engine today:

```
NameError: name 'set_timeout' is not defined
```

## Why this one is a bit different from CSE/PVML

This is a *runtime* error, not a compile-time one — the Resolver accepts `set_timeout` fine (it's in the shared MISC group), so the compiled WASM module builds successfully; it just has no binding for the name at execution time. The gap is in `src/engines/wasm/library.ts`'s `miscLib` — the WASM engine's own per-function codegen table (separate from `MiscBuiltins`, which CSE uses directly, and separate from PVML's own primitive table) — which doesn't have an entry for `set_timeout`/`clear_all_timeout` at all yet.

## The re-entry problem, WASM's version

Same underlying issue as CSE and PVML: `f` must still be callable when the real timer fires, potentially after the WASM instance's original exported run has already returned. Two things worth investigating when picking this up:

1. **Whether the WASM engine's existing JSPI-based async machinery (used for Source Academy module calls — see `src/engines/wasm/index.ts` and the language-directory's WASM "Chrome-only, JSPI" note) is reusable here, or a different shape of problem.** JSPI as used for module calls lets a synchronous-looking WASM→JS call suspend on a JS Promise, resumed later — but that's WASM code *calling out* and waiting. `set_timeout` needs the opposite direction: JS code (a real `setTimeout` firing) needs to call *back into* the WASM instance's closure-invocation machinery later, potentially after the instance's original top-level run has already completed. Worth checking whether a WASM instance's exported functions stay callable after that point the way a compiled subroutine would, or whether this needs its own re-entry design closer to CSE's problem than PVML's.
2. **Whether this should even be JSPI-gated / Chrome-only**, given `set_timeout` itself doesn't obviously need JSPI the way a synchronous-looking module call does — a plain host import calling back into an already-instantiated module's exports might not need it at all, which would make this *less* browser-restricted than the module-call path, not equally restricted.

## Reference

[...]

Issue #320, opened 2026-07-22
PVML: set_timeout/clear_all_timeout support


martin-henz:
## Background

#311 added `set_timeout(f, t)` / `clear_all_timeout()` to the MISC stdlib group (needed for the tone matrix quest, Q5B), implemented for py2js only so far — see that PR. Calling `set_timeout` under PVML today:

```
RunError: Primitive function set_timeout not implemented
```

## Why this should be the easiest of the three follow-ups

Unlike CSE (see the separate CSE issue), PVML already has `invokeValueAsync` — built specifically for sound_matrix's own `set_timeout(f, t)` use case, i.e. exactly this "call a PVML closure later, from outside the normal step loop" problem has already been solved once for a module-gated version of the same feature. This issue is about making that a first-class *language* capability instead of one gated behind importing a specific module, matching how #311 phrased it for py2js.

## Two things worth checking when picking this up

1. **Where the "not implemented" error actually comes from.** It's not `src/stdlib/misc.ts`'s CSE stub — PVML has its own separate primitive dispatch table (distinct from the `MiscBuiltins`/`GroupName.MISC` builtins CSE and the Resolver both key off), and `set_timeout`/`clear_all_timeout` were never registered there. `set_timeout` currently passes validation (it's in the shared MISC group, so the Resolver accepts it everywhere) but has no PVML-side primitive backing it — find where sound_matrix's own `set_timeout` is wired in and use the same registration path.
2. **Whether `invokeValueAsync` can be reused as-is** or needs generalizing — sound_matrix's usage may assume things (e.g. being called from within a module context) that don't hold for a bare language-level `set_timeout`.

## Reference

`src/engines/py2js/runtime.ts`'s `set_timeout`/`clear_all_timeout` builtins (and `src/tests/py2js-set-timeout.test.ts`) are the working reference implementation and test shape — same signature/semantics (delay in milliseconds, `clear_all_timeout` cancels every pending callback scoped to the current run), just needing PVML's own re-entry path instead of py2js's "already a JS closure" shortcut.

Issue #319, opened 2026-07-22
CSE: set_timeout/clear_all_timeout support


martin-henz:
## Background

#311 added `set_timeout(f, t)` / `clear_all_timeout()` to the MISC stdlib group (needed for the tone matrix quest, Q5B), implemented for py2js only so far — see that PR. The CSE machine currently has an explicit stub in `src/stdlib/misc.ts`:

```ts
static set_timeout(...): NoneValue {
  throw new Error("set_timeout is not yet supported by the CSE machine");
}
```

## Why this is the hardest of the three

`f` must still be callable when the real timer fires, which may be well after the `evaluate()` run that scheduled it has already finished and torn down — `context.control`/`context.stash` are gone by then. Unlike:
- **py2js**, where a compiled Python function already *is* a plain JS closure — a real `setTimeout` callback just calls it directly on the JS event loop, no interpreter involved (see `src/engines/py2js/runtime.ts`'s `set_timeout` builtin and its doc comment).
- **PVML**, which already has `invokeValueAsync` (built for sound_matrix's own `set_timeout`) — see the separate PVML issue.

...the CSE machine is a step-based interpreter (control/stash/environment in `src/engines/cse/`) with no existing way to resume evaluation of a closure call "from cold," outside its own step loop.

## What picking this up would involve

1. Some mechanism to re-enter the CSE machine's evaluate loop given just a closure value and call arguments — not from an in-progress `Control`/`Stash`, but fresh, as if starting a new top-level call.
2. Output routing: `context.streams.stdout` (`src/engines/cse/streams.ts`) needs to still be valid (or be re-established) when the callback fires after the original `evaluateChunk` has already resolved — mirrors the problem py2js's implementation solved via `onOutput`, but CSE's streams are tied to the conductor/context lifecycle differently.
3. Error handling: an error raised inside a fired callback can no longer reach the chunk that scheduled it (same reasoning as py2js's implementation, adapted to CSE's error-reporting path — `displayError`/`ErrorType`).
4. `clear_all_timeout()` needs the same per-run timer bookkeeping py2js's `pendingTimeouts` provides, scoped appropriately for CSE's context lifecycle.

## Reference

`src/engines/py2js/runtime.ts`'s `set_timeout`/`clear_all_timeout` builtins (and `src/tests/py2js-set-timeout.test.ts`) are the working reference implementation and test shape to match, adapted to CSE's very different execution model.

PR #322, opened 2026-07-22
Add set_timeout/clear_all_timeout to MISC (py2js)

### LINKED ISSUE #311: Add set_timeout and clear_all_timeout to MISC
This is needed for the tone matrix quest: Q5B

martin-henz:
## Summary
Closes #311. Moves sound_matrix's own `set_timeout(f, t)`/`clear_all_timeout()` into the language itself (MISC stdlib group) — same signature and units — rather than keeping it gated behind importing that module. Needed for the tone matrix quest (Q5B).

- **py2js**: fully implemented. A compiled Python function already is a plain JS closure, so a real `setTimeout` callback just calls it directly on the event loop — no interpreter re-entry needed. Scheduled via `rt.acall` (not the sync path) so a callback that itself needs a frontend round-trip (e.g. calling an imported module function) still works. An error raised inside a fired callback is reported through `onOutput` rather than silently lost, since the chunk that scheduled it has already resolved by the time it fires. `clear_all_timeout` is scoped per-`Py2JsRuntime`.
- **CSE / PVML / WASM**: not implemented yet. Each fails its own way rather than crashing — CSE has an explicit stub, PVML raises a "primitive not implemented" `RunError`, WASM raises a runtime `NameError` (compiles fine since the name is in the shared MISC group, just has no codegen binding). Tracked as separate follow-ups:
  - #319 — CSE (hardest: no existing cold re-entry path for a closure call)
  - #320 — PVML (likely easiest: `invokeValueAsync` already exists, built for sound_matrix's own use of this)
  - #321 — WASM (different shape of gap: runtime NameError, not compile-time; open question on whether it even needs the JSPI/Chrome-only path)

## Test plan
- [x] `src/tests/py2js-set-timeout.test.ts` — 8 new tests: fire-after-delay, ordering across multiple timeouts, `clear_all_timeout` cancellation, error propagation from a fired callback, argument validation, and a callback that itself calls into an imported module function
- [x] `stdlib-conformance-py2js.test.ts` updated to skip `set_timeout`/`clear_all_timeout` in the CPython/CSE cross-check (permanent, expected mismatch — the CSE reference always throws "not yet supported")
- [x] Full suite: 61 suites, 19,237 tests passing, 0 regressions
- [x] `tsc --noEmit`, `eslint` clean


martin-henz:
@coderabbitai review

PR #326, opened 2026-07-23
Remove dead RunOptions.envSteps parameter

### LINKED ISSUE #274: runner.ts: RunOptions.envSteps is dead code, silently does nothing
## Summary

`src/runner.ts`'s `RunOptions.envSteps` (doc'd as "Maximum number of environment steps before stopping. Default: 100000.") is accepted by `runCode()` and threaded all the way down to `generateCSEMachineStateStream`, but has no effect at all. Only `stepLimit` (default `-1`, i.e. unlimited) actually stops execution.

## Root cause

In `src/engines/cse/interpreter.ts`, `generateCSEMachineStateStream`'s own parameter is named `_envSteps` (underscore-prefixed, i.e. deliberately unused), and its only use site is commented out:

```ts
export async function* generateCSEMachineStateStream(
  code: string,
  context: Context,
  control: Control,
  stash: Stash,
  _envSteps: number,
  stepLimit: number,
  ...
) {
  ...
  while (command) {
    // Return to capture a snapshot of the control and stash after the target step count is reached
    // if (!isPrelude && steps === envSteps) {
    //   yield { stash, control, steps }
    //   return
    // }

    // Step limit reached, stop further evaluation
    if (!isPrelude && steps === stepLimit) {
      ...
      handleRuntimeError(context, new error.StepLimitExceededError(source, stmtOrExpr));
    }
    ...
```

Practical consequence: `runCode()` called with default `RunOptions` has **no step limit whatsoever** — `stepLimit` defaults to `-1` (unlimited) and `envSteps` (default 100000) is a no-op. A genuine infinite loop (`while True: x = x + 1`) passed to `runCode()` with default options hangs forever; only explicitly passing `stepLimit` actually bounds it. Found while writing `src/tests/runner.test.ts` — a test that tried to rely on `envSteps: 50` to cap a runaway loop hung indefinitely; switching it to `stepLimit: 50` worked as expected.

## Possible fixes

- If `envSteps` was meant as a real safety limit distinct from `stepLimit` (e.g. counting only environment-changing steps, for snapshot-pagination purposes in the visualizer), re-enable the commented-out check, or
- If `envSteps` is now redundant with `stepLimit` and its meaning was superseded, drop it from `RunOptions`/`generateCSEMachineStateStream`'s signature entirely rather than keeping a parameter that silently does nothing, and update `RunOptions`'s own doc comment (currently actively misleading: "Default: 100000" implies a limit is being applied).

Either way, `runCode()`'s headless callers (the CLI, and anything else built on `src/runner.ts`) currently have no real protection against a runaway user program unless they explicitly pass `stepLimit`.

martin-henz:
## Summary
Closes #274.

`envSteps` was threaded from `RunOptions` (`runner.ts`) all the way down to `generateCSEMachineStateStream` (`engines/cse/interpreter.ts`), but its only use site was commented out — the parameter (underscore-prefixed there, `_envSteps`) did nothing. `runCode()` called with default options had no real step protection at all: `stepLimit` defaults to `-1` (unlimited) and `envSteps` (documented default 100000) was a no-op, so a genuine infinite loop would hang forever unless a caller explicitly passed `stepLimit`.

## Why removed rather than re-enabled

`generateCSEMachineStateStream`'s `while` loop already `yield`s unconditionally on every single step. The commented-out `envSteps` check was meant to yield-and-*return early*, pausing execution after N steps — but nothing needs that today. `PyCseMachinePlugin.ts`'s `collectSnapshots` — the actual live consumer backing the CSE-machine visualizer, not just `runner.ts`'s headless path — already does its own truncation via `maxSnapshots`:
```ts
if (snapshots.length >= maxSnapshots) break;
```
It consumes the full unconditional-yield stream and stops itself once it has enough. Whatever pagination need `envSteps` was originally meant to serve is already solved a different way, at the consumer rather than the producer — re-enabling the dead check would have been redundant with logic that already works correctly, not a fix for a real gap.

## What changed
Removed the same dead parameter from every layer it was threaded through — there were two entirely separate options-threading paths carrying it, not just one:
- `RunOptions` (`runner.ts`) and its two `collectSnapshots` call sites
- `IOptions` (`engines/cse/interpreter.ts`) — a second, separate layer for `evaluate()`/`runCSEMachine`, easy to miss if only searching from `runner.ts`
- `collectSnapshots`'s own signature (`PyCseMachinePlugin.ts`) and its `generateCSEMachineStateStream` call, plus the two hardcoded-`100000` call sites in `PyCseEvaluator.ts` (the real production conductor evaluator)
- An orphaned `envSteps: -1` field in `context.ts`'s `createEmptyRuntime()` that wasn't even part of `Context.runtime`'s declared type and was read nowhere — same abandoned feature, a separate leftover fragment of it

## Test plan
- [x] Full suite: 60 suites, 19,228 tests, 0 regressions
- [x] `tsc --noEmit`, `eslint` clean


Akshay-2007-1 (approved):
`experiments/py2js/bench.ts:104` and `experiments/py2js/compare.ts:126` still call `runCode(code, 1, { envSteps: 1_000_000_000 })`. These are TypeScript object literals passed directly as the options argument, so with envSteps gone from RunOptions this would now fail TypeScript's excess-property check if type-checked. It wasn't caught by tsc --noEmit/CI because experiments/ is excluded.

LGTM otherwise!

PR #332, opened 2026-07-23
WASM: implement str() and repr(), fix print()'s float formatting

### LINKED ISSUE #323: WASM: repr() doesn't return a string — implemented as a print() copy
## Repro

```python
print(repr(1))
```
Under the WASM engine, this prints **two** lines: `1` and `None`. And:
```python
print("[" + repr(1) + "]")
```
throws `TypeError: unsupported operand type(s) for this arithmetic operation`.

Isolated against the same call shape with `str()` instead of `repr()` — `print("[" + str(1) + "]")` works fine — so this is specific to `repr()`, not string concatenation in general.

## Root cause

`src/engines/wasm/library.ts`:
```ts
funcHelper("print", 1, true).body(x => wasm.call(LOG_FX).args(x)),
funcHelper("repr", 1, true).body(x => wasm.call(LOG_FX).args(x)),
```

`repr` is registered with the exact same body as `print` — it calls the logging function as a side effect and returns whatever that returns (apparently `None`), rather than converting its argument to a string and returning that string. So:
- `repr(1)` alone prints `1` as an (unwanted) side effect, then the caller sees a `None` return value — hence the double print above.
- `"[" + repr(1) + "]"` is actually `"[" + None + "]"`, which fails as a type error on the `+`.

## Impact

This is what actually broke the original `llist_to_string(llist(1, 2))` repro brought up while checking #258 — `_llist_to_string` (`src/stdlib/linked-list.prelude.ts`) calls `repr(xs)` and concatenates its result into the box-and-pointer string it builds. Any prelude function relying on `repr()`'s return value (not just for its printed side effect) will hit this on WASM. Not related to #258 — that issue is scoped to PVML-in-browser; this is WASM-specific, in a completely separate codegen table (`miscLib` in `library.ts`, distinct from `MiscBuiltins` which CSE uses directly).

## Fix direction

`repr` needs its own WASM implementation that produces and returns a string value (matching `MiscBuiltins.repr` in `src/stdlib/misc.ts`'s semantics — CPython-style `repr()`, e.g. quoting strings, distinguishing `int`/`float`/`bool`/`None` rendering), not a call to the print/log host import. Worth checking whether `str()`'s WASM implementation (which does work per the isolation above) can be reused/adapted, since `repr()` and `str()` differ only for a handful of types (strings and a few container reprs) in this dialect.

martin-henz:
## Summary

Closes #323, which was originally filed as "repr() returns None instead of a string" but turned out to be larger in scope on re-investigation: `str()` didn't exist on WASM at all (`NameError: name 'str' is not defined`), and `repr()` was literally a copy of `print()`'s body — same void return, no string produced.

- Added `TO_STR_FX`/`TO_REPR_FX` (new WASM functions, `src/engines/wasm/runtime/stdlib.ts`) that reuse `LOG_FX`'s existing type-dispatch formatting by calling back into the exported `log()` and popping the rendered text — the same trick `log_list` already uses for its element rendering. The formatted text is then allocated as a real WASM string via the same malloc/encode/write/`makeString` sequence the `tokenize` host import uses (factored out into a shared `allocateWasmString` helper in `hostImports.ts`, also now used by `tokenize` itself instead of its own inline copy).
- Wired `str`/`repr` into `src/engines/wasm/library.ts`'s `miscLib`, replacing the broken `repr` entry.
- `repr()` applies Python's exact string-quoting rules (single-quote-preferring, switches to double quotes when the string itself contains a `'`) by reusing `escape()` from `src/stdlib/utils.ts` (now exported) instead of reinventing it.

### Two bugs found and fixed along the way

**`log_float` formatting.** `wasm.import("console", "log").func("$_log_int")` and `wasm.import("console", "log").func("$_log_float")` declared the *same* `(module, field)` import key, so both silently resolved to the same JS `console.log` at instantiation regardless of which local alias a call site referenced. `$_log_float` ended up bound to the int-formatting implementation (`value.toString()`), so `print(5.0)` rendered `"5"` instead of `"5.0"`, and large floats never got exponential notation. Renamed the imports to distinct `log_int`/`log_float` fields and gave `log_float` its own implementation using `toPythonFloat` (the same function CSE's own `toPythonString` uses, now exported from `stdlib/utils.ts`), so WASM's float formatting now matches CSE exactly.

**Shadow-stack leak (first implementation attempt).** `GET_LEX_ADDR_FX` already pushes a fetched argument onto the shadow stack when it's GC'able (protecting it during any subsequent allocation before it's consumed). The first version of `TO_STR_FX`/`TO_REPR_FX` didn't account for this: it read the argument, computed a result, and returned a *new* string (itself pushed by `MAKE_STRING_FX` as a side effect) — leaving the *original* argument's shadow-stack entry never popped. That stray entry sat underneath the new result and threw off whatever GC'able-operand accounting the surrounding expression did next. Concretely, `"[" + repr("hi") + "]"` produced garbage (unrelated static string-table bytes) instead of `"['hi']"`, because the `+` operator's string-concat branch unconditionally pops exactly two shadow-stack entries for its two operands and got the wrong ones. Fixed by popping the input's shadow-stack entry (if GC'able) before computing, mirroring the existing `IS_INT_FX`/`IS_PAIR_FX`/`MAKE_PAIR_FX` per-argument convention.

### Test changes

- `src/tests/wasm/wasm-misc.spec.ts` — new `str() and repr()` describe block: per-type formatting, quoting rules, and a regression test for the shadow-stack leak (string concatenation of a `str()`/`repr()` result).
- `src/tests/operator-conformance-wasm.test.ts` — dropped the numeric-tolerance workaround for float results (previously needed only because of the `log_float` bug above); floats are now compared as exact text like every other type. Verified with a full 3933-case run.
- `src/tests/wasm-modules.test.ts` — updated 3 expectations that had baked in the old float-formatting bug: imported numeric module values (`DataType.NUMBER`) map to Python floats at the WASM boundary (see `moduleInterop.ts`), so `print()` on an imported `42` now correctly renders `"42.0"`, not `"42"`.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `yarn lint` (targeted eslint on changed files)
- [x] `yarn test` — full suite, twice, both green (19234 passed / 3166 skipped)
- [x] `operator-conformance-wasm.test.ts` run standalone (3933 cases) with the tightened exact-text float comparison


martin-henz:
@coderabbitai review

martin-henz:
@coderabbitai review

PR #340, opened 2026-07-23
CSE machine: label loop-increment assigns and tag expression statements

### LINKED ISSUE #270: CSE machine: indicate expression statements
Currently we have:

Image

followed by

Image

It's not clear where the "pop" is coming from. 

I'm suggesting that the first "1 + 2" should indicate that it appears as a statement, maybe with a color say orange for expression statements. Then when an orange expression is on the control, it will turn into a pop followed by the expression in white.

The overall "notional machine" principle: The machine should be predictable: I should be able to predict the next step from the current step. Not making expression statements explicit violates this principle.

### LINKED ISSUE #293: CSE machine visualization: assignment to loop variable currently too opaque
Currently:

Image

Intead of 'assign' the control item should be 'i = i'

Akshay-2007-1:
## Summary
- \`generateForIncrement\`'s synthetic per-iteration Assign node (the loop variable update) had synthetic tokens, so \`serializeControlItem\` fell back to the generic \`"assign"\` KIND_LABEL instead of showing which variable was set to what. It now carries a \`syntheticLabel\` of \`"x = <value>"\` (fixes #293).
- Tagged \`SimpleExpr\` control items with \`tag: "expressionStatement"\` so the visualizer can render them distinctly. Currently an expression statement like \`1 + 2\` is indistinguishable from a plain sub-expression until a "pop" mysteriously appears on a later step to discard its value (fixes #270). The frontend consumes this tag to render the coloring — see source-academy/frontend#4159.

image

## Test plan
- [x] \`jest src/tests/PyCseMachinePlugin.test.ts\` — added unit tests for the tag and an end-to-end regression test asserting a \`for\` loop's control stack shows \`"x = 0"\` instead of \`"assign"\`; all 91 tests pass.
- [x] Manually verify in the Python playground: stepping through a \`for\` loop shows \`x = <value>\` instead of \`assign\`, and an expression statement's control item is visually distinct (color change lands via the frontend PR above).

### DIFF
--- a/src/conductor/plugins/PyCseMachinePlugin.ts
+++ b/src/conductor/plugins/PyCseMachinePlugin.ts
@@ -330,7 +330,16 @@ function serializeControlItem(item: ControlStackItem, code: string): SerializedI
       nodeMeta.bodyNodeTypes = bodyArray.map(n => PY_TO_JS_NODE_TYPE[n.kind] ?? "Identifier");
     }
 
-    return Object.keys(nodeMeta).length > 0 ? { displayText, metadata: nodeMeta } : { displayText };
+    // Expression statements ("1 + 2" as a bare statement) are indistinguishable from a
+    // plain sub-expression once pushed — the machine only reveals it was a statement
+    // later, when the "pop" instruction that discards its value appears (issue #270).
+    // Tag it so the visualizer can mark it distinctly (e.g. a different color) up front.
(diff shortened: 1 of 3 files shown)


Akshay-2007-1:
@coderabbitai review

martin-henz (approved):
LGTM

PR #353, opened 2026-07-25
py2js: closure_call_sync fast path for module closures called from Python

martin-henz:
## Summary

`moduleInterop.ts`'s `moduleToPython` builds a `PyFunction` for every module closure Python imports (`from math import sqrt`, or a hot per-call accessor like pix_n_flix's proposed `get_pixel_value`). Until now this was always `asyncOnly`: a defensive-backstop sync body that unconditionally threw, and an async body that unconditionally paid the mandatory `AsyncGenerator` tax via `closure_call_unchecked` — even when the closure carries a `.sync` twin proving it never needs a real host round-trip.

That fast path already exists and is already generic: `GenericDataHandler.closure_call_sync` looks up a closure's underlying function and calls its `.sync` twin directly if present, regardless of whether the closure came from a module or from Python — it's the exact mechanism `pythonToModule`'s own `.sync` attach already uses for the opposite direction (a Python closure crossing into a module, e.g. sound's wave sampling, 44,100 calls/sec). Only `moduleToPython` never attempted it.

## What changed

- `moduleToPython`'s `DataType.CLOSURE` case now tries `dh.closure_call_sync` first, converting arguments/result through the same restricted scalar converters (`pythonToModuleSync`/`moduleToPythonSync`) the other direction's fast path already uses. Falls back to the same `"needs a frontend round-trip"` error only when no `.sync` twin is available *for that specific call* — a dynamic, per-call decision now, rather than a fixed `asyncOnly` flag rejecting every sync call unconditionally regardless of what the closure was actually capable of.
- `asyncBody` (the always-correct fallback used on the async spine, `acall`) is unchanged.
- Updated `py2js-module-interop.test.ts`: the old "CLOSURE becomes an asyncOnly PyFunction" test is renamed to reflect that this is now a per-call outcome, not a static flag, and a new test confirms a closure carrying `.sync` is now callable synchronously with no round-trip.

## Motivation

Prompted by planning the `pix_n_flix` module's migration to Conductor ([source-academy/modules](https://github.com/source-academy/modules)). A redesigned pixel API (`get_pixel_value`/`set_pixel_value` over an opaque image handle, instead of decomposing every video frame into a Source-level array) needs to be called from inside a student's filter at high frequency — up to `width * height * 8` times per frame. That's the same shape of hot loop as sound's per-sample wave callback, just the reverse direction across the module boundary, and this engine README's own benchmark puts the difference at ~15-20ns/call (sync) vs. ~390-420ns/call (mandatory async-generator) — the gap between a usable frame budget and roughly 2.5fps at 400x300.

Depends on **source-academy/conductor#54**, which fixes `BaseModulePlugin.initialise()` silently dropping a module method's `.sync` twin across `Function.prototype.bind()` — without that fix, a module's `.sync` twin would never survive registration for this change to actually see.

## Test plan

- [x] `yarn jest src/tests/py2js` — 214/214 passing (15 suites)
- [x] `yarn jest src/tests/operator-conformance-py2js.test.ts src/tests/stdlib-conformance-py2js.test.ts` — 4785/4785 passing (no regressions in general execution semantics)
- [x] `npx tsc --noEmit` — clean
- [x] `npx eslint` on both changed files — clean

### DIFF
--- a/src/engines/py2js/moduleInterop.ts
+++ b/src/engines/py2js/moduleInterop.ts
@@ -66,12 +66,18 @@ import { Py2JsRuntime, Py2JsRuntimeError, PyOpaque, PyValue } from "./runtime";
  * Synchronous, scalar-only counterparts to moduleToPython/pythonToModule -
  * used only by the `.sync` fast path a Python closure crossing into a module
  * gets below (see GenericDataHandler.closure_call_sync's doc for the overall
- * design). Cover exactly the value shapes a scalar-in/scalar-out closure (a
- * wave function, sampled 44100x/sec by the sound module) needs: numbers,
- * booleans, strings, None. Return `undefined` for anything else (pairs,
- * closures, opaques, complex) - the "no sync path" signal, safe to use for
- * *arguments* (nothing has run yet) but not for the *result* once the real
- * call has already happened - see pyClosureFunc.sync below.
+ * design). Cover the value shapes a scalar-in/scalar-out closure (a wave
+ * function, sampled 44100x/sec by the sound module) needs - numbers,
+ * booleans, strings, None - plus OPAQUE, needed for an opaque-handle
+ * accessor (pix_n_flix's get_pixel_value/set_pixel_value, taking an opaque
+ * image handle as `source`/`dest`, sampled up to width*height*8x/frame).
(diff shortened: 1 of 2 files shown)


martin-henz:
@coderabbitai review

PR #362, opened 2026-07-26
Fix jsdoc build failure from duplicate declarations in misc.js

Akshay-2007-1:
## Summary
While checking that #361 actually made `set_timeout`/`clear_all_timeout` show up on https://docs.sourceacademy.org/python/, I found the entire MISC section renders as empty ("No names predeclared in MISC") — and same for python_1 through python_4, which all bundle `misc.js`.

Root cause: `docs/lib/misc.js` declares `print`, `error`, and `time_time` twice each (identical doc content both times — looks like a merge artifact). `jsdoc`'s parser rejects duplicate top-level function declarations as a syntax error, so it has been failing to parse `misc.js` on every single build:

```
ERROR: Unable to parse .../docs/lib/misc.js: Identifier 'error' has already been declared. (85:9)
```

`scripts/jsdoc.sh`'s `run_jsdoc` doesn't check `jsdoc`'s exit code, so `deploy-docs.yml` has been reporting green while quietly shipping an empty MISC page for a while.

- Removes the three duplicate declarations, keeping one copy of each (content was byte-identical between copies).
- No content is lost — every function that was in either copy is still documented.

## Test plan
- [x] Ran the same `jsdoc` generator/config as `scripts/jsdoc.sh` (`run_jsdoc`-equivalent invocation) against `docs/lib/misc.js` before/after: before, it errors and produces an empty page; after, it parses clean and every MISC function (including `print`, `error`, `time_time`, `set_timeout`, `clear_all_timeout`) renders.
- [x] Also ran the full python_4 group build (misc.js + math.js + linked_list.js + pairmutator.js + list.js + stream.js + mce.js) with no errors.

### DIFF
--- a/docs/lib/misc.js
+++ b/docs/lib/misc.js
@@ -74,17 +74,6 @@ defaults to <CODE>0</CODE>.
 function complex(v = 0, i = 0) { }
 
 /**
- * Prints the provided <CODE>*object</CODE> arguments to the standard output (similar to a simplified
- * <CODE>print</CODE>) and then raises an exception. This function accepts a variable number of arguments,
- * converts them to their <CODE>string</CODE> representations using <CODE>str()</CODE>, outputs them (with a
- * newline) just like what <CODE>print</CODE> does, and immediately halts execution by raising an exception.
- *
- * @param {any} <CODE>*object</CODE> - Objects to be printed to the standard output
- * @returns {NoneType} the <CODE>None</CODE> value
- */
-function error(...object) { }
-
-/**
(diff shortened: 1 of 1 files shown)


Akshay-2007-1:
@coderabbitai, are these changes justified?

martin-henz:
Oh, yes, I also noticed the empty sections today but forgot to make an issue. Thanks for fixing this @Akshay-2007-1 .

martin-henz (approved):
LGTM

PR #375, opened 2026-07-31
Fix Py2JS stack overflow on cyclic list print/str/repr

### LINKED ISSUE #341: Py2JS: print of cyclic data structure gives stack overflow
Consider the following snippet : 
```python
a = [1]
a[0] = a
print(a)
```

This same code gives a bug called `too much recursion` in Py2JS. Also try:

```python
a = [1]
a[0] = a
```

Image

martin-henz:
## Summary

Fixes #341. `print`, `str()`, and `repr()` in the Py2JS engine crashed with `RangeError: Maximum call stack size exceeded` on a self-referential list (e.g. `a = [1]; a[0] = a`), instead of printing CPython's cycle marker.

- **Root cause**: two separate PyValue→CSE-Value conversion paths — `toDisplayValue` (`runtime.ts`, used by `print`) and `toTagged`/`toTaggedList` (`stdlibBridge.ts`, used by every generically-bridged builtin including `str`/`repr`) — rebuilt a fresh `Value` tree on every call with no cycle awareness, so a self-referential JS array recursed forever *before* the shared `stringify()` renderer's own, already-correct ancestor-cycle detection ever got a chance to run.
- **Fix**: both conversions now memoize by object identity, allocating each list's placeholder *before* descending into its elements/spine, so a cyclic `PyValue` converts into a genuinely cyclic `Value` graph — the same shape `stringify()` already knows how to render safely. `toTaggedList`'s iterative long-list optimization (avoiding one JS stack frame per element for long proper lists) is preserved.
- Also updates `stringify.ts`'s cycle marker text from `"...<circular>"` to CPython's actual `"[...]"`, verified against real CPython's `Py_ReprEnter` behavior (each nesting level of a cycle gets its own bracket pair, e.g. `[[...]]` for direct self-reference). This is a shared file also used by the CSE machine, so its cyclic-print output becomes CPython-accurate too — no existing tests depended on the old text.

## Test plan

- [x] Added `describe("printing a self-referential list mirrors CPython's cycle marker...")` in `src/tests/py2js-lists.test.ts`, covering: cycle at top level, cycle at list position `[1]`, cycle at position `[2]`, cycle nested one level down (not at top level), cycle nested two levels down, indirect/mutual cycles between two lists, and a shared-but-acyclic control case (must NOT trigger the marker) — each cross-checked against the CSE machine (`runCode`) and manually verified against real CPython.
- [x] Added `str()`/`repr()` cases confirming they no longer overflow either.
- [x] Full existing test suite passes (19,466 tests, no regressions), including the stdlib/operator conformance sweeps that already exercise `str`/`repr` over the full type universe.
- [x] `tsc --noEmit` clean.

### DIFF
--- a/src/engines/py2js/runtime.ts
+++ b/src/engines/py2js/runtime.ts
@@ -52,7 +52,7 @@
  */
 import { DataType, TypedValue } from "@sourceacademy/conductor/types";
 import { numericCompare, pythonMod } from "../cse/utils";
-import type { Value } from "../cse/stash";
+import type { ListValue, Value } from "../cse/stash";
 import { toPythonFloat, toPythonString } from "../../stdlib/utils";
 import { PyComplexNumber } from "../../types";
 import { stringify } from "../../utils/stringify";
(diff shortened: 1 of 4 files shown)


martin-henz:
@coderabbitai reviews

PR #376, opened 2026-07-31
Recognize breakpoint() with any number of arguments as a breakpoint

### LINKED ISSUE #257: breakpoint() with args is silently ignored by breakpoint-navigation in both the Stepper and the CSE Machine
## Summary

Both the substitution stepper's breakpoint navigation (#197) and the CSE Machine's breakpoint navigation (#256) only recognize a **zero-argument** call to `breakpoint()` as a navigable breakpoint marker:

- Stepper: `reduce.ts`'s `stepHead` (`ExpressionStatement` case) — detects a no-arg call resolving to the builtin `breakpoint`.
- CSE Machine: `interpreter.ts`'s `generateCSEMachineStateStream` — checks `command.instrType === InstrType.APPLICATION && command.numOfArgs === 0`.

In both cases, calling `breakpoint()` with arguments (e.g. `breakpoint(5)`) still **evaluates without error** (it's a no-op builtin that ignores whatever args it's given either way), but the call is silently **not** recorded as a breakpoint step — so the double-chevron navigation just skips over it, with no error, warning, or other indication to the student that their breakpoint wasn't actually registered.

## Why this matters

Real Python's `breakpoint()` does accept arguments (forwarded to `sys.breakpointhook`), so a student writing `breakpoint(5)` isn't doing anything syntactically wrong from a "real Python" perspective — but in py-slang's stepper/CSE Machine, it silently loses the one behavior (navigation) that makes `breakpoint()` useful pedagogically here. This is a confusing failure mode: the code runs fine, nothing looks wrong, but the debugging aid just doesn't work.

## Proposed direction (needs discussion, not decided here)

Make `breakpoint()` strictly zero-arg in py-slang, and raise a proper error (e.g. a `TypeError`-style arity error, consistent with how other builtins validate via `@Validate`) when called with any arguments — rather than silently accepting-but-ignoring them. This would apply consistently to both the Stepper and the CSE Machine, since they currently share the same "no-op no matter the args" builtin implementation semantics.

This needs a design discussion with @Shrey5132 (author of #197, the original stepper breakpoint() support) before implementing, since:
- It changes runtime behavior for existing student code that may already call `breakpoint()` with args (even though it's a no-op today).
- It should land consistently across both interpreters rather than as a one-off fix in either.

## References

- Stepper support: #197 (issue #188)
- CSE Machine support: #256 (issue #189)

martin-henz:
## Summary

Fixes #257. Real Python's `breakpoint()` signature is `breakpoint(*args, **kws)` — it accepts any number of positional (and keyword) arguments, forwarded to `sys.breakpointhook`. Both the CSE machine's and the substitution stepper's own `breakpoint` builtin already reflected this (no arity validation at all, args silently ignored either way — see `stdlib/misc.ts` and `conductor/stepper/builtins.ts`).

The bug was purely in **breakpoint-navigation detection**: both engines only recognized a *zero-argument* call as a navigable breakpoint marker, so `breakpoint(5)` ran fine but the double-chevron navigation silently skipped it, with no error or indication to the student.

This takes the direction of matching CPython's actual signature (accept any arity, same as the argument-count-agnostic no-op `breakpoint` already does) rather than the issue's alternate "make it strictly zero-arg and raise a TypeError otherwise" suggestion — since real Python code calling `breakpoint(5)` is not doing anything wrong, it should just work, navigation included.

- **CSE machine** (`interpreter.ts`): the zero-arg restriction existed only because the callee sits on top of the stash at the point checked, which is true only when there are zero args (APPLICATION pops `numOfArgs` args before popping the callee). Added `Stack.peekAt(depthFromTop)` and used `peekAt(numOfArgs)` to reach the callee regardless of arity.
- **Stepper** (`reduce.ts`): generalized the `arguments.length === 0` check to `arguments.every(isValue)` — the same "is this argument already a value" predicate `contractCall` uses to decide when a call is ready to apply. A still-reducible argument (e.g. `breakpoint(1 + 2)`) now steps through normally first; the statement is flagged as the breakpoint-navigation target once every argument has become a value, mirroring the point the CSE machine's APPLICATION instruction would actually fire.

## Test plan

- [x] `PyCseMachinePlugin.test.ts`: `breakpoint()` with one/several/computed positional arguments, and via an alias, all still record a `breakpointSteps` entry; confirms it still evaluates to `None` and ignores its arguments.
- [x] `getSteps.test.ts` (stepper): `breakpoint(5)` / `breakpoint(1, 2, 3)` still evaluate as a no-op and are flagged as the "Evaluating breakpoint statement" debugger target; `breakpoint(1 + 2)` confirms the argument is reduced to a value (`3`) *before* the statement gets flagged, not on first encounter; aliasing (`bp = breakpoint; bp(1, 2)`) still works.
- [x] Full existing test suite passes (19,485 tests, no regressions).
- [x] `tsc --noEmit` clean.

### DIFF
--- a/src/conductor/stepper/__tests__/getSteps.test.ts
+++ b/src/conductor/stepper/__tests__/getSteps.test.ts
@@ -1622,6 +1622,49 @@ describe("Python stepper — breakpoint() marks a debugger breakpoint (#188)", (
     );
     expect(flagged).toHaveLength(0);
   });
+
+  // Real Python's breakpoint(*args, **kws) takes any number of arguments (forwarded to
+  // sys.breakpointhook); the stepper's own no-op breakpoint entry already ignores them regardless
+  // (see builtins.ts), so the bare-statement form should still be recognised as a debugger target
+  // for any arity, not just zero (issue #257).
+  test.each([
+    ["one literal argument", "breakpoint(5)"],
+    ["several literal arguments", "breakpoint(1, 2, 3)"],
+  ])("%s: still evaluates as a no-op and is flagged as a breakpoint target", (_desc, call) => {
+    expect(result(`${call}\n1 + 1`)).toBe("2");
+    const flagged = steps(`${call}\n1 + 1`).filter(
+      step => step.markers?.[0]?.redexNodeType === "DebuggerStatement",
+    );
+    expect(flagged).toHaveLength(1);
+    expect(flagged[0].markers?.[0]?.explanation).toBe("Evaluating breakpoint statement");
+  });
+
+  test("an argument that is itself reducible is stepped through first, then flagged once all arguments are values", () => {
+    // breakpoint(1 + 2): the argument isn't a value yet, so the statement must not be flagged as the
+    // debugger target on the very first encounter — only once `1 + 2` has itself reduced to `3` (the
+    // point analogous to the CSE machine's APPLICATION instruction actually firing).
+    const src = "breakpoint(1 + 2)\n1 + 1";
+    const flagged = steps(src).filter(
+      step => step.markers?.[0]?.redexNodeType === "DebuggerStatement",
+    );
+    expect(flagged).toHaveLength(1);
+    expect(flagged[0].markers?.[0]?.explanation).toBe("Evaluating breakpoint statement");
+    // Confirm the argument really was reduced before the flagged step: the flagged step's own tree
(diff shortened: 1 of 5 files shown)


martin-henz:
@coderabbitai review

martin-henz:
@coderabbitai review

PR #377, opened 2026-07-31
Fix NameReassignmentError's location to point at the original declaration

### LINKED ISSUE #211: NameReassignmentError says "already declared here" instead of "reassigned here"
```python
x = 1
x = 2
```
raises `NameReassignmentError`, but the message says "already declared here" instead of "reassigned here".

The actual location pointed to is ok.

martin-henz:
## Summary

Fixes #211. `NameReassignmentError`'s "already declared here" pointer was supposed to point at a name's *original* declaration, but instead pointed at whatever the resolver's hoisting pass happened to visit last for that name — sometimes the reassignment statement that's already flagged by the error's own primary location, sometimes a completely unrelated *later* occurrence.

**Root cause**: `env.names` is used for two purposes that don't compose. `Resolver.resolve()` hoists every `Assign`/`FunctionDef` in a statement list before validating any of it (so forward references within the same block resolve), calling `Environment.declareName()` unconditionally — last write wins, with no "first wins" rule. That's correct for "is this name bound in this scope," but `no-reassignment.ts`'s validator and `functionVarConstraint` both also read `env.names.get(name)` to report *where the name was first declared*, which by then has already been overwritten.

Concretely, for `x = 1; x = 2; x = 3`, both the `x = 2` and `x = 3` errors' "already declared here" pointed at line 3 (the last hoisted occurrence) — for the `x = 2` error, that's a location *after* the error itself, which is incoherent. A same-scope-reassignment case with only two statements happened to self-reference (pointing at the very statement already flagged), masking how wrong the general case was.

**Fix**: `Environment` gains `firstDeclarations`, a separate map seeded from the environment's initial `names` (so a function/lambda parameter's own token is correctly its own "first declared" location) and updated by `declareName` only the first time a name is set in that scope, never after. Both throw sites that report a reassignment's original location — `no-reassignment.ts`'s validator and `functionVarConstraint` (the same bug, reached via a different path: same local assigned in both arms of an `if`/`else`, or a nested function/lambda body shadowing an enclosing local) — now read from `firstDeclarations` instead of `names`.

## Test plan

- [x] Added a `resolver.test.ts` describe block covering: two reassignments (points at line 1, not line 2), three reassignments (still points at line 1, not the last-hoisted line 3), a parameter reassigned in its own function body (points at the `def` line, not the reassignment), and the same local assigned in both if/else arms (points at the if-arm, not the else-arm where the error fires).
- [x] Full existing test suite passes (19,479 tests, no regressions) — nothing asserted on the old (wrong) location text.
- [x] `tsc --noEmit` clean.

### DIFF
--- a/src/resolver/resolver.ts
+++ b/src/resolver/resolver.ts
@@ -21,6 +21,20 @@ export class Environment {
   // The parent of this environment
   enclosing: Environment | null;
   names: Map<string, Token>;
+  /**
+   * Where each name in `names` was *first* declared in this scope — for diagnostics only (e.g.
+   * NameReassignmentError's "already declared here" pointer). `names` itself gets overwritten on
+   * every hoist of a name (declareName has no "first wins" rule: `resolve(Stmt[])`'s hoisting pass
+   * runs unconditionally over every Assign/FunctionDef in a statement list, including a list
+   * re-hoisted more than once against the same environment, e.g. an if-arm and its else-arm, which
+   * share their enclosing scope's environment — see visitIfStmt), which is required for forward
+   * references to resolve but means `names.get(name)` can end up pointing at an unrelated later
+   * occurrence, or even the very statement the reassignment error is itself about, rather than the
+   * name's true original declaration (issue #211). Seeded from the environment's initial `names`
+   * (so a function/lambda parameter's own token is its own "first declared" location — see the
+   * constructor), then updated by declareName only the first time a name is set, never after.
+   */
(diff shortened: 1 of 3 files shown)


martin-henz:
@coderabbitai review

martin-henz:
Actually the fix is much simpler: "already declared here" should say "reassigned here" 
The error is: reassignment is not allowed in Python §1 and §2. So the issue to be pointed out to the programmer is the reassignment, not the original declaration.

PR #379, opened 2026-08-01
Add local-file ("folder") imports for py2js (§2+)

### LINKED ISSUE #378: Copy folder treatment from js-slang
js-slang implements "folders" (see frontend and js-slang) in an interesting way, harking back to JavaScript's IEFE-based implementation of import/export.

Can we copy this for py-slang?

martin-henz:
## Summary

Copies js-slang's own "folder" treatment — local, relative imports (`from .foo import bar`) resolved against a program's own multi-file map — scoped to the py2js engine. Closes #378.

- **No conductor changes needed.** Local file resolution goes through `IRunnerPlugin.requestFile`, which already exists in the published `@sourceacademy/conductor` package and is already backed by the frontend's folder-mode file store (`BasicEvaluator.startEvaluator` already uses it to fetch the entrypoint). `Py2JsEvaluatorBase` just passes `path => conductor.requestFile(path)` down.
- **A genuine source-to-source transform** (`src/modules/localImports.ts`, engine-agnostic — not tied to py2js internals): every locally-imported file becomes a `def __module_x__(): ...` wrapping that file's own statements, returning a real SICPy linked list of `(name, value)` pairs; the importer's `from .foo import bar` becomes a plain lookup against it via a generated `__access_named_export__` walk — line for line the same shape as js-slang's own local-import prelude (`localImport.prelude.ts`). The whole multi-file program flattens into one ordinary Python source string *before* anything is parsed or compiled, so:
  - it's exactly the artifact a student can be shown via `print(__program__)`,
  - it needs zero engine-specific runtime support (no new opcodes, no new native helpers beyond what the generated Python itself calls: `pair`/`head`/`tail`, already in the linked-list stdlib group).
- **Requires SICPy §2 or higher**, mirroring js-slang's own restriction to Source §2+ for the identical reason — the pair/list transfer structure doesn't exist below that chapter. CSE, PVML, and the WASM engine now reject a relative import explicitly (clear error) rather than silently mistreating a dotted path as a Source Academy module name — full support for those engines is out of scope here.
- **Also fixes a real, separate bug this surfaced**: `Py2JsEvaluatorBase` never supplied `__program__` at all, so referencing it crashed *every* chapter with a raw `ReferenceError` rather than working or failing gracefully. `Py2JsSession` now keeps `__program__` current itself, automatically, from whatever text it's actually compiling for a given chunk — mirroring how the CSE machine already handles this internally — rather than relying on an external caller to remember to wire it up.

## Test plan
- [x] `npm test` — full suite green (19,499 passed, 3,166 pre-existing skips, 0 failures)
- [x] New unit tests for the bundler itself (`src/tests/local-imports-bundler.test.ts`): two-file imports, nested/dotted paths, aliasing, diamond dependencies (shared file runs exactly once), circular imports, missing file/name, chapter-1 rejection, cross-chunk caching via `priorGlobalNames`
- [x] Engine-level tests (`src/tests/py2js-local-imports.test.ts`) via the real `runCodePy2JsDual`/`Py2JsSession` API, plus a `fileGetter` (conductor.requestFile shape) test and a `__program__`-shows-the-flattened-text test
- [x] `__program__` regression tests added/fixed (`src/tests/py2js-interpreter-support.test.ts`)
- [x] Verified live end-to-end in the real Source Academy frontend (Folder mode) against a locally-built `Py2JsEvaluator{1,2,3,4}.js`, unmodified conductor, and unmodified frontend: a two-file folder-mode program correctly imports a function and a value from a sibling file at §2/§3/§4, and is correctly rejected at §1

https://claude.ai/code/session_011b8eJsqoDGiuesf5uQ42oG

### DIFF
--- a/src/ast-types.ts
+++ b/src/ast-types.ts
@@ -365,15 +365,18 @@ export namespace StmtNS {
     readonly kind = "FromImport";
     module: Token;
     names: { name: Token; alias: Token | null }[];
+    level: number;
     constructor(
       startToken: Token,
       endToken: Token,
       module: Token,
       names: { name: Token; alias: Token | null }[],
+      level: number,
     ) {
       super(startToken, endToken);
       this.module = module;
       this.names = names;
+      this.level = level;
     }
     override accept(visitor: Visitor<any>): any {
       return visitor.visitFromImportStmt(this);
--- a/src/conductor/Py2JsEvaluator.ts
+++ b/src/conductor/Py2JsEvaluator.ts
@@ -77,9 +77,35 @@ abstract class Py2JsEvaluatorBase extends BasicEvaluator {
       onOutput: line => this.conductor.sendOutput(line),
       onPendingWorkChange: delta => (delta > 0 ? this.beginPendingWork() : this.endPendingWork()),
       requestInput: prompt => this.conductor.requestInput(prompt),
+      // A local import (`from .foo import x`, source-academy/py-slang#378)
+      // resolves a sibling file through this — conductor's own
+      // requestFile(fileName), already backed by whatever multi-file store
+      // the host has (e.g. the frontend's folder-mode BrowserFS), the same
+      // primitive BasicEvaluator.startEvaluator already uses to fetch the
+      // entrypoint itself. No new host-side protocol needed.
+      fileGetter: path => this.conductor.requestFile(path),
       dataHandler,
       dataVisualizer: this.dataVisualizerPlugin,
     });
+  }
+
+  /**
+   * The host's own entrypoint file name (e.g. "/main.py" in a folder-mode
+   * program) — BasicEvaluator's default evaluateFile discards it, but a
+   * local import needs to know what directory "." means, so this override
+   * captures it on the session before delegating exactly as the default
+   * implementation would.
+   *
+   * (`__program__` needs no help from here: Py2JsSession.runChunkInternal
+   * sets it itself, automatically, from whatever text it's actually about
+   * to compile — see that method's own comment. This evaluator used to
+   * never supply it at all, which crashed every reference to `__program__`
+   * with a raw ReferenceError rather than a Python-level NameError; fixed
+   * at the source rather than by adding another external call here.)
+   */
+  async evaluateFile(fileName: string, fileContent: string): Promise<void> {
+    this.session.setEntrypointFilePath(fileName);
+    return this.evaluateChunk(fileContent);
   }
 
   async evaluateChunk(chunk: string): Promise<void> {
--- a/src/conductor/PyPvmlEvaluator.ts
+++ b/src/conductor/PyPvmlEvaluator.ts
@@ -134,6 +134,14 @@ abstract class PyPvmlEvaluatorBase extends BasicEvaluator {
     const importsByModule = new Map<string, { name: string; alias: string | undefined }[]>();
     for (const stmt of ast.statements) {
       if (stmt instanceof StmtNS.FromImport) {
+        if (stmt.level > 0) {
+          // Local-file imports (leading dots) aren't implemented on PVML yet
+          // (see py2js for the supported engine) — reject explicitly rather
+          // than treating the dotted path as a conductor module name.
+          throw new Error(
+            "ImportError: relative imports (e.g. 'from .module import name') are not yet supported by this engine.",
+          );
+        }
         const moduleName = stmt.module.lexeme;
         if (!importsByModule.has(moduleName)) {
           importsByModule.set(moduleName, []);
--- a/src/conductor/PyPvmlEvaluator.ts
+++ b/src/conductor/PyPvmlEvaluator.ts
@@ -5,6 +5,7 @@ import { moduleToPvml } from "../engines/pvml/modules";
 import { PVMLBoxType } from "../engines/pvml/types";
 import { PVMLCompiler } from "../engines/pvml/pvml-compiler";
 import { PVMLInterpreter } from "../engines/pvml/pvml-interpreter";
+import { RELATIVE_IMPORT_NOT_SUPPORTED_MESSAGE } from "../errors";
 import { parse } from "../parser/parser-adapter";
 import { analyzeWithEnvironments } from "../resolver";
 import linkedList from "../stdlib/linked-list";
@@ -138,9 +139,7 @@ abstract class PyPvmlEvaluatorBase extends BasicEvaluator {
           // Local-file imports (leading dots) aren't implemented on PVML yet
           // (see py2js for the supported engine) — reject explicitly rather
           // than treating the dotted path as a conductor module name.
-          throw new Error(
-            "ImportError: relative imports (e.g. 'from .module import name') are not yet supported by this engine.",
-          );
+          throw new Error(RELATIVE_IMPORT_NOT_SUPPORTED_MESSAGE);
         }
         const moduleName = stmt.module.lexeme;
         if (!importsByModule.has(moduleName)) {
--- a/src/conductor/PyWasmEvaluator.ts
+++ b/src/conductor/PyWasmEvaluator.ts
@@ -68,6 +68,15 @@ class PyWasmEvaluator extends BasicEvaluator {
     const importsByModule = new Map<string, { name: string; alias: string | undefined }[]>();
     for (const stmt of ast.statements) {
       if (stmt instanceof StmtNS.FromImport) {
+        if (stmt.level > 0) {
+          // Local-file imports (leading dots) aren't implemented on the WASM
+          // engine yet (see py2js for the supported engine) — reject
+          // explicitly rather than treating the dotted path as a conductor
+          // module name.
+          throw new Error(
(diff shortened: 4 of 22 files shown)


martin-henz:
@coderabbitai review

martin-henz:
## Review

### Overview

This PR adds local-file ("folder") imports to the **py2js** engine: `from .foo import bar` / `from ..pkg.foo import bar` now resolve against a program's own file map instead of a conductor module, gated to SICPy §2+ (needs `pair`/`head`/`tail`). The implementation is a genuine **source-to-source bundler** (`src/modules/localImports.ts`): it flattens an entrypoint plus every file it transitively imports into one Python source string — each dependency becomes a `def __module_X__(): ...; return <pair-list of its own top-level names>`, and the importer's `from .foo import bar` becomes `bar = __access_named_export__(__exports_foo__, "bar")`. That flattened text is then parsed and run through the **completely unmodified** existing single-file pipeline. This mirrors js-slang's own local-module preprocessor and needs zero new engine runtime state and zero conductor protocol changes (conductor's existing `requestFile` already supplies the file map). CSE/pvml/wasm get an explicit rejection guard so `.foo` isn't silently misread as a literal conductor-module name. A related, previously-broken `__program__` (predeclared name showing the compiled program text) is fixed as a side effect, made fully automatic inside `Py2JsSession`.

Overall this is a well-designed, well-documented feature with strong architectural restraint (no new runtime primitives, reuses `priorGlobalNames`/deterministic naming for cross-chunk caching instead of adding a cache). I found two concrete correctness bugs while reviewing (details below, both reproduced locally), plus several smaller gaps.

### Code quality & style

- `src/modules/localImports.ts` is thoroughly documented — the module doc comment explains the design rationale (js-slang parity, the "no export keyword" simplification, the deterministic-naming cache mechanism) well enough that a future maintainer doesn't have to reverse-engineer *why*.
- The old `asModule` compile option, `exportsStmt` codegen, and the runtime-level `localModuleExports`/`accessExport` cache were all cleanly removed rather than left as dead code from an earlier design iteration — good cleanup.
- Minor duplication: the "relative imports not yet supported by this engine" message is copy-pasted as a literal string across `PyPvmlEvaluator.ts`, `PyWasmEvaluator.ts`, `pvml-compiler.ts`, and `builderGenerator.ts` (CSE's version differs slightly, wrapped in a proper `RuntimeSourceError` subclass). Worth a shared constant, and worth deciding whether PVML/WASM should also use a typed error rather than a bare `Error` (see "Potential issues" below).
- `resolveLocalModulePath`'s `".."`-segment-popping branch is unreachable through the current grammar (`dotted_name` can only ever produce identifier segments, never a literal `..`) — harmless as defensive code for an exported utility function, but worth a one-line comment noting it's speculative/untested-by-design so a future reader doesn't wonder if it's dead code by mistake.

### Specific suggestions

1. **Fix `slugify`'s collision** (see Issue 1 below) — likely just needs a length-prefixed or otherwise unambiguous encoding of path segments rather than blanket non-alphanumeric-to-`_` collapsing.
2. **Fix `indentBlock`'s naive line-based reindentation** (see Issue 2 below) — it needs to be indentation-aware of string-literal spans (or reindentation could be avoided entirely by wrapping via `exec`/a different structuring approach that doesn't require shifting the dependency's own source text).
3. Add tests exercising the CSE/PVML/WASM rejection guards directly — currently zero tests confirm any of the 4 new guard call sites actually fire (see Test coverage below).
4. Consider having `runCodePy2Js` (the sync, one-shot path) throw a clear "local imports require the dual/async API" error immediately if `files`/`fileGetter` is passed, or if the parsed program contains a `level > 0` import — right now this silently no-ops and the caller instead gets a confusing internal "no pending import value" `SystemError` at runtime.
5. Minor: `prepareDual` calls `bundleLocalImports(...)` without a `priorGlobalNames` argument (defaults to empty), while `Py2JsSession.runChunkInternal` passes `new Set(Object.keys(this.rt.globals))`. Doesn't affect correctness today (each `prepareDual` call gets a fresh runtime), but it's an asymmetry between the two call sites worth a comment if intentional.

### Potential issues / risks

**1. Correctness — `slugify()` can collide two different files onto the same generated name.**
`slugify` collapses any run of non-alphanumeric characters to a single `_`:
```ts
function slugify(path: string): string {
  return path.replace(/^\/+/, "").replace(/[^a-zA-Z0-9]+/g, "_");
}
```
`/pkg/utils.py` and `/pkg_utils.py` both slugify to `pkg_utils_py` (verified directly). Since `dotted_name` only allows identifier characters, **both** are reachable via ordinary import syntax in the same program — `from .pkg.utils import x` and `from .pkg_utils import x` — so a program that imports from both a `pkg/utils.py` file *and* a sibling `pkg_utils.py` file would silently generate two colliding `def __module_pkg_utils_py__()` / `__exports_pkg_utils_py__` bindings, with the second overwriting/shadowing the first. Narrow in practice, but a real, silent-failure-mode bug rather than a clean error. A prefix-free encoding (e.g. joining segments with a rare separator before collapsing, or hashing the full path) would close this.

**2. Correctness — bundling corrupts multi-line triple-quoted strings in dependency files.**
`indentBlock` re-indents a statement's copied source text by naively adding 4 spaces to *every* line:
```ts
function indentBlock(text: string, indent: string = INDENT): string {
  return text.split("\n").map(line => (line.length > 0 ? indent + line : line)).join("\n");
}
```
The tokenizer's triple-quote regexes (`src/parser/lexer.ts`) do span newlines (`"""(?:[^\\]|\\.)*?""""`), so a dependency file containing a multi-line triple-quoted string has its literal content mutated by this reindentation. I reproduced this directly:
```python
# /utils.py
doc = """
line one
line two
"""
```
Original string value: `"\nline one\nline two\n"`. After being bundled as a dependency and actually executed: `"\n    line one\n    line two\n    \n"` — 4 spaces injected into every interior line, plus a stray trailing line. This only affects **non-entrypoint** files (their source text gets wrapped in `def __module_X__(): ...` and thus reindented); the entrypoint's own body is spliced in unindented. Worth a targeted fix or at minimum an explicit test documenting the limitation if it's deemed out of scope for this PR.

**3. Error-type inconsistency across the four new engine guards.**
CSE's guard raises `RelativeImportNotSupportedError extends RuntimeSourceError`, giving it the engine's normal source-location/formatting treatment. PVML's and WASM's guards throw a bare `new Error(...)` instead, losing that location/formatting fidelity relative to how those engines otherwise surface errors. Consider aligning these, even if it's just wrapping in each engine's own existing error base class.

**4. Missing test coverage for the guard call sites.** `PyPvmlEvaluator.ts`, `PyWasmEvaluator.ts`, `pvml-compiler.ts`, `builderGenerator.ts`, and `cse/interpreter.ts` all gained a `level > 0` rejection check, but none of them have a test confirming the rejection actually fires (vs., say, silently falling through to the old "treat `.foo` as literal module name `foo`" behavior). This is the one place in the PR where behavior changed without corresponding test coverage — everything in `src/modules/localImports.ts` and the py2js integration is otherwise very thoroughly tested.

**5. Level-beyond-root relative imports clamp instead of raising the specific CPython error.** `resolveLocalModulePath("/main.py", 5, "utils")` clamps to `/utils.py` rather than CPython's `ImportError: attempted relative import beyond top-level package`. Not wrong — a genuinely-missing file still errors — but the error message a student would see ("module not found") is less specific than it could be. Low priority.

**6. Sequential (not parallel) file resolution.** `bundleFile`'s dependency loop `await`s each `fileGetter(path)` call one at a time rather than resolving independent sibling imports concurrently. This is the right call for correctness (shared `BundleState` mutation isn't safely parallelizable as written), but for a `conductor.requestFile` backed by a real round-trip, a program with many sibling imports pays for them serially. Likely negligible for typical folder-sized student programs; flagging in case it becomes a real cost.

### Test coverage

Very strong overall: `local-imports-bundler.test.ts` and `py2js-local-imports.test.ts` between them cover two-file imports, nested/dotted paths, aliasing, diamond dependencies (asserting exactly one def/invocation), transitive imports, circular-import rejection, missing-file rejection, the "undefined name → `None`, not an error" parity choice, chapter-1 rejection, level-0 hoisting/ordering, cross-chunk caching via `priorGlobalNames`, and the async `fileGetter` (conductor-shape) path including `setEntrypointFilePath`. The `__program__` fix also has dedicated new/rewritten tests. The gaps are the ones called out above: the four engine-guard call sites, and no test with a multi-line string literal or with two files whose paths collide under `slugify`.

### Security

No injection risk from import target names — they're always grammar-restricted to identifier characters before being interpolated into generated string literals (`JSON.stringify(spec.name.lexeme)`), so there's no way to break out of the generated string literal. The one real concern is issue 2 above (source text reindentation), which is a correctness bug, not an injection vector — it can corrupt a string's *value* but not the surrounding code structure.

### Performance

The chosen design (bundle once into flat text, then run the existing unmodified single-file pipeline) is the cheapest possible integration — no added engine overhead beyond the one bundling pass. Cross-chunk caching by deterministic name avoids redundant re-bundling/re-execution across REPL chunks. The only note is the sequential fetch behavior in issue 6.

PR #384, opened 2026-08-02
Wire gutter-click breakpoints to their closest enclosing statement

### LINKED ISSUE #383: Proper breakpoints with gutter clicks
Currently gutter clicks (red dots) are already supported in the frontend (ace editor) and recorded.

Also both stepper and CSE machine support the function breakpoint() and provide support to jump to the next breakpoint.

Let's wire the gutter clicks properly so that they refer to the closest surrounding statement and mark the statement as having a breakpoint.

The stepper and CSE machine can then check if a given statement has a breakpoint and register the evaluation accordingly.

[Note that this improves over the treatment of gutter clicks in js-slang, which literally inserts "debugger;" statements, which caused various issues there.]

martin-henz:
## Summary

- Closes #383. A gutter click only carries a line number; `markBreakpoints` (`src/breakpoints.ts`) resolves it to the closest enclosing statement (innermost covering statement, or the nearest following one for a blank/comment line) and flags it with a bolted-on `hasBreakpoint` property, mirroring `engines/cse/types.ts`'s existing `isEnvDependent`.
- The CSE machine (`engines/cse/interpreter.ts`) and the stepper (`conductor/stepper/reduce.ts`) each check that flag right next to their existing `breakpoint()` detection and record the same kind of step — so a gutter click reuses the host's existing breakpoint-navigation controls (the "jump to next breakpoint" double-arrow) rather than needing a new wire message.
- The host serves the clicked line numbers through the existing `/__cse_config__` virtual file (extended with a new `breakpointLines` field), fetched via the shared `src/conductor/runConfig.ts` helper — the same channel `stepLimit` already used, so no changes are needed to `@sourceacademy/conductor` or the other vendored protocol packages. Both `PyCseEvaluator.ts` and `PyStepperEvaluator.ts` (which previously fetched no config at all) now read it.
- Along the way, fixed a real bug caught by the stepper's own tests: a naive implementation fired the breakpoint marker on *every* intermediate step of a multi-step statement (e.g. each reduction step of a multi-term RHS expression), not just the first time execution reaches that line. Fixed in `stepHead` (`reduce.ts`) by clearing the flag on the carried-forward copy once it's been reported once.

## Test plan

- [x] `npx jest` — full suite passes (74/74 suites, 19554 tests)
- [x] `npx tsc --noEmit` — clean
- [x] `npx eslint src` — no new errors/warnings
- New/extended coverage:
  - `src/tests/breakpoints.test.ts` — unit tests for the line-to-statement resolution algorithm
  - `src/tests/PyCseMachinePlugin.test.ts` — gutter breakpoints in loops, functions, if-headers, blank lines, past-end-of-file, composed with an explicit `breakpoint()` call
  - `src/conductor/stepper/__tests__/getSteps.test.ts` — same coverage for the stepper, plus a regression test asserting a multi-step statement is only flagged once
  - `src/tests/PyCseEvaluator.test.ts` — end-to-end test mocking `requestFile("/__cse_config__")`

https://claude.ai/code/session_01MvqZCkjN8CLSBWCoNSZbAG

### DIFF
--- a/src/breakpoints.ts
+++ b/src/breakpoints.ts
@@ -1,0 +1,84 @@
+/**
+ * Resolves gutter-clicked line numbers to AST nodes, for issue #383: "Proper breakpoints with
+ * gutter clicks". A gutter click only carries a line number, not an AST node, so this walks the
+ * parsed tree once per run and flags the statement each requested line resolves to. The stepper
+ * (`conductor/stepper`) and CSE machine (`engines/cse`) each separately check the resulting flag
+ * and treat it exactly like an explicit `breakpoint()` call.
+ */
+
+import { StmtNS } from "./ast-types";
+
+/** Bolted onto statement instances, mirroring `engines/cse/types.ts`'s `isEnvDependent` — not
+ * part of the generated `ast-types.ts`, since it's per-run state, not part of the grammar. */
+type BreakpointFlag = { hasBreakpoint?: boolean };
+
+/** Every statement in `stmt`'s subtree (including `stmt` itself), depth-first. Only the five
+ * statement kinds that own a nested statement list (`ast-types.ts`) are recursed into. */
+function collectStatements(stmt: StmtNS.Stmt, out: StmtNS.Stmt[]): void {
+  out.push(stmt);
+  switch (stmt.kind) {
+    case "If": {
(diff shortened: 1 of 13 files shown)


martin-henz:
@coderabbitai review

PR #386, opened 2026-08-03
Stepper: support a subset of module imports

### LINKED ISSUE #385: Stepper: can't handle import statements
The CSE machine handles import statements by treating imported functions as builtins. The stepper just gives "evaluation stuck" when it encounters import.

martin-henz:
## Summary

Closes #385. `from X import Y` used to make the stepper report "Evaluation stuck" the instant it saw the import line, regardless of whether the imported name was ever used. This wires up real module resolution for the substitution stepper, mirroring the CSE machine's own two-phase (load-then-run) model — for a deliberately scoped subset of modules (see below).

- `FromImport` now translates to a no-op statement instead of an inert placeholder identifier, and renders the student's actual import text (`from rune import circle`, not `pass`).
- New `OpaqueValue`/`ModuleFunction` `StepNode` kinds represent opaque module handles (e.g. a `rune` `Rune`) and imported callables as legitimate, substitutable stepper values.
- `ModuleLoaderRunnerPlugin` is registered on `PyStepperEvaluatorBase`, exactly mirroring how `PyCseEvaluatorBase`/`Py2JsEvaluatorBase`/`PyPvmlEvaluatorBase` already do it.
- New `moduleInterop.ts` resolves and substitutes a program's imports before stepping begins (`resolveImports`), and calls imported functions (`callModuleFunction`) — trying the synchronous `closure_call_sync` fast path first, falling back to draining the mandatory async `closure_call_unchecked` generator.
- **Scope**: a Python closure (a `lambda`/`def`, or a bare reference to a static built-in) passed as an *argument* into a module call is declined — the call simply stays irreducible ("Evaluation stuck"), the same honest degrade `input()`/`time_time()` already get. Supporting that direction would mean the module calling back into Python mid-call, which needs the reducer to re-enter its own step machinery — out of scope for this pass. This covers modules like `rune` (pure data/opaque handles in, opaque handles out) but not ones like `sound` that take a Python function as an argument.
- A relative import (`from .foo import x`) is now a clear, student-actionable error (mirroring the CSE machine's `RelativeImportNotSupportedError`) rather than silently rendering as inert text.
- `reduce.ts`/`getSteps.ts` are now `async` end-to-end to support this — the biggest mechanical part of the diff, since it also required migrating the existing (large) stepper test suite off synchronous helper calls.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx eslint` clean on all touched files
- [x] `npx prettier --list-different` clean
- [x] Full test suite green (`npx jest`): 19565 passed, 0 failed
- [x] New unit tests for the `FromImport` → no-op translation, the relative-import error, and aliased-import display text
- [x] New integration tests against a fake module (the same `GenericDataHandler` + `ModuleLoaderRunnerPlugin` harness `py2js-from-import.test.ts` uses), not just the no-evaluator-wired degrade path: a constant export substituted as a literal, a function export actually called, an opaque return value (both created-and-discarded and used/rendered), a declined Python-closure argument staying stuck, a missing module, and a name a module doesn't export

https://claude.ai/code/session_017w4FqCep4j8XyFpF4kpVue

### DIFF
--- a/src/conductor/PyStepperEvaluator.ts
+++ b/src/conductor/PyStepperEvaluator.ts
@@ -2,9 +2,11 @@ import { STEPPER_DIRECTORY_ID } from "@sourceacademy/common-stepper";
 import { ConductorError, EvaluatorSyntaxError } from "@sourceacademy/conductor/common";
 import { BasicEvaluator, type IRunnerPlugin } from "@sourceacademy/conductor/runner";
 import { RunnerStatus } from "@sourceacademy/conductor/types";
+import { ModuleLoaderRunnerPlugin } from "@sourceacademy/runner-module-loader";
 
 import { markBreakpoints } from "../breakpoints";
 import { parse } from "../parser";
+import { asInterfacableEvaluator, GenericDataHandler } from "./GenericDataHandler";
 import { registerAutoCompletePlugin } from "./plugins/autocomplete";
 import { fetchRunConfig } from "./runConfig";
 import { evaluatePython } from "./stepper/getSteps";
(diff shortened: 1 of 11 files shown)


martin-henz:
@coderabbitai review

martin-henz:
@coderabbitai review

martin-henz:
Follow-up: this PR merged just before CodeRabbit's second review landed. That review's one finding (a concern about `substitute`'s `clone()` corrupting `ModuleFunction.closure` during AST substitution) turned out to be a false positive — see #387 for the investigation and a regression test pinning it down.

Issue #181, opened 2026-06-29
CSE machine: use/assignment before global/nonlocal declaration


martin-henz:
```
def f():
    x = 1
    global x
    return x

f()
print(x)
```
should produce
```
  File "/Users/henz/Downloads/test.py", line 3
    global x
    ^^^^^^^^
SyntaxError: name 'x' is assigned to before global declaration
```
but succeeds, in py-slang.

Simliar: use of variable before global, assignment before nonlocal, and use of variable before nonlocal.

Note that Python uses static (textual) order in the function body to determine which is before what. So these are static checks, not runtime checks.

My hunch is that we should implement this carefully. Otherwise, we get undesired noise in CS1101S created by students who want to prove that our Python implementation is not proper (for the Python sublanguages we define).

Issue #179, opened 2026-06-29
CSE machine: nonlocal and parameter


martin-henz:
```
x = 0
def f(y):
    x = 1
    def g(x):
       nonlocal x
f(2)
```
should produce
```
  File "/Users/henz/Downloads/test.py", line 7
    nonlocal x
    ^^^^^^^^^^
SyntaxError: name 'x' is parameter and nonlocal
```
but succeeds in py-slang.

Issue #178, opened 2026-06-29
CSE machine: nonlocal and global error


martin-henz:
```
x = 0
def f(y):
    x = 1
    def g(z):
      global x
       nonlocal x
f(2)
```
should produce
```
  File "/Users/henz/Downloads/test.py", line 7
    global x
    ^^^^^^^^
SyntaxError: name 'x' is nonlocal and global
```
but succeeds on py-slang.

PR #388, opened 2026-08-03
Match CPython's wording for unresolvable nonlocal declarations

### LINKED ISSUE #187: nonlocal resolution errors don't match CPython's wording (generic NameNotFoundError instead of specific message)
Related to #186 (error disparity for py-slang versus Python) — this is the same class of issue, scoped to `nonlocal` resolution errors specifically. Found while testing PR #182 locally against real CPython output (cc @martin-henz).

## The problem

When `nonlocal x` can't be resolved, py-slang always falls back to the generic resolver message:

```
NameNotFoundError: This name is not found in the current or enclosing environment(s).
Perhaps you meant to type '<suggestion>'?
```

CPython gives a specific, more diagnostic message for this exact situation:

```
SyntaxError: no binding for nonlocal 'x' found
```

Two reproductions:

### 1. No binding for the name anywhere in an enclosing function
```python
def outer():
    def inner():
        nonlocal x
        x = 1
    inner()
outer()
```
- **py-slang**: `NameNotFoundError: This name is not found in the current or enclosing environment(s).`
- **CPython**: `SyntaxError: no binding for nonlocal 'x' found`

### 2. Binding exists further out, but an intermediate function explicitly declares the same name `global` (which correctly blocks `nonlocal` from skipping past it)
```python
def grandouter():
    x = 1
    def outer():
        global x
        x = 99
        def inner():
            nonlocal x
            x = 2
        inner()
    outer()
    return x
print(grandouter())
```
- **py-slang**: `NameNotFoundError at line 7 ... nonlocal x ... This name is not found in the current or enclosing environment(s). Perhaps you meant to type 'x'?`
- **CPython**: `SyntaxError: no binding for nonlocal 'x' found`

Note the `Perhaps you meant to type 'x'?` suggestion in case 2 — the Levenshtein suggestion logic found some other binding named `x` in scope (the very thing blocking resolution) and suggested it back, which is actively confusing since the identifier being resolved is already `x`. The suggestion algorithm probably shouldn't offer the exact same name as a "did you mean" correction.

Both cases are *correctly rejected* (still a `SyntaxError` under the hood, since `NameNotFoundError extends BaseResolverError extends SyntaxError`) — this is purely a message-wording/diagnostics-quality issue, not a correctness bug.

## Suggested fix

In `visitNonLocalStmt` (`src/resolver/resolver.ts`), give the "not found" fallback a dedicated message ("no binding for nonlocal '<name>' found") instead of reusing the generic `NameNotFoundError`, and suppress/adjust the suggestion when the closest match is the identifier itself.

martin-henz:
## Summary

Closes #187.

`visitNonLocalStmt` fell back to the generic `NameNotFoundError` when a `nonlocal x` declaration has no binding in any enclosing function scope. CPython raises a specific `SyntaxError` for this exact diagnostic:

```
SyntaxError: no binding for nonlocal 'x' found
```

The generic fallback also produced a misleading `Perhaps you meant to type 'x'?` suggestion in the issue's second repro case — where a same-named binding exists in scope but is unreachable (an intermediate function declares `global x`, correctly blocking `nonlocal` from skipping past it). The Levenshtein-closest match ends up being the identifier itself, since a name reaching this point is by definition already spelled correctly — there's simply no binding it can resolve to as `nonlocal`.

## Fix

- Reuse `ScopeConflictError` (`src/resolver/errors.ts`) — already used for the sibling #178-#181 nonlocal/global scope-conflict diagnostics, and whose reported error name is literally `"SyntaxError"`, matching CPython's own exception class here — instead of introducing a new error class.
- Never call `suggestName` for this case, since CPython never offers a "did you mean" for it either.
- `visitGlobalStmt` was checked and doesn't have the analogous bug — `global x` is unconditionally accepted by the resolver (real Python only raises at runtime if `x` is never actually assigned, which py-slang already defers to) — so it's untouched.

## Test plan

- [x] Tightened the two existing repro-case tests in `src/tests/nonlocal.test.ts` (previously only asserted `.toThrow(SyntaxError)`) to assert the exact CPython-matching message and the absence of the misleading suggestion.
- [x] Updated `src/tests/wasm/wasm-functions.spec.ts`'s "cannot declare parameter as nonlocal" test: it was incidentally also asserting a `NameNotFoundError` was present, since the same name failed both the #179 parameter/nonlocal conflict and (as an unrelated side effect) the "no binding" check; now asserts the more accurate `ScopeConflictError`.
- [x] `npx tsc --noEmit` clean
- [x] `npx eslint`/`npx prettier --list-different` clean
- [x] Full suite green: 19568 passed, 0 failed

https://claude.ai/code/session_017w4FqCep4j8XyFpF4kpVue

### DIFF
--- a/src/resolver/resolver.ts
+++ b/src/resolver/resolver.ts
@@ -876,14 +876,20 @@ export class Resolver implements StmtNS.Visitor<void>, ExprNS.Visitor<void> {
     // skipping the current function itself (nonlocal can never bind to it).
     const found = this.hasEnclosingFunctionBinding(name, this.functionDefStack.length - 2);
     if (!found) {
+      // A dedicated message matching CPython's own wording for this exact diagnostic (#187) —
+      // not the generic NameNotFoundError, whose "Perhaps you meant to type 'x'?" suggestion is
+      // actively misleading here: the closest Levenshtein match is often `name` itself (e.g. the
+      // very binding an intermediate `global name` is blocking resolution past), since a name
+      // reaching this point is, by definition, spelled correctly and simply unresolvable as
+      // nonlocal — there is nothing sensible to suggest instead.
       this.errors.push(
-        new ResolverErrors.NameNotFoundError(
+        new ResolverErrors.ScopeConflictError(
           stmt.name.line,
           stmt.name.col,
           this.source,
           stmt.name.indexInSource,
-          stmt.name.indexInSource + stmt.name.lexeme.length,
-          this.environment?.suggestName(stmt.name) ?? null,
+          stmt.name.indexInSource + name.length,
+          `no binding for nonlocal '${name}' found`,
         ),
       );
     }
--- a/src/tests/nonlocal.test.ts
+++ b/src/tests/nonlocal.test.ts
@@ -835,9 +835,16 @@ grandouter()
     // resolve to as nonlocal), no "Perhaps you meant to type 'x'?" suggestion should appear
     // either, even though `x` genuinely exists in scope here (the very global-shadowed binding
     // blocking resolution) and would otherwise be the Levenshtein-closest — and thus most
-    // confusing possible — suggestion.
-    expect(() => toPythonAstAndResolve(code, 3)).toThrow(/no binding for nonlocal 'x' found/);
-    expect(() => toPythonAstAndResolve(code, 3)).not.toThrow(/Perhaps you meant/);
+    // confusing possible — suggestion. Resolved once, not twice: both checks read the same
+    // thrown error's message.
+    let error: Error | undefined;
+    try {
+      toPythonAstAndResolve(code, 3);
+    } catch (e) {
+      error = e as Error;
+    }
+    expect(error?.message).toMatch(/no binding for nonlocal 'x' found/);
+    expect(error?.message).not.toMatch(/Perhaps you meant/);
   });
 
   test("nonlocal with no binding construct anywhere in any enclosing function is still rejected", () => {
@@ -849,7 +856,13 @@ def outer():
     inner()
 outer()
 `;
-    expect(() => toPythonAstAndResolve(code, 3)).toThrow(/no binding for nonlocal 'x' found/);
-    expect(() => toPythonAstAndResolve(code, 3)).not.toThrow(/Perhaps you meant/);
+    let error: Error | undefined;
+    try {
+      toPythonAstAndResolve(code, 3);
+    } catch (e) {
+      error = e as Error;
+    }
+    expect(error?.message).toMatch(/no binding for nonlocal 'x' found/);
+    expect(error?.message).not.toMatch(/Perhaps you meant/);
   });
 });
--- a/src/tests/nonlocal.test.ts
+++ b/src/tests/nonlocal.test.ts
@@ -830,7 +830,14 @@ def grandouter():
     outer()
 grandouter()
 `;
-    expect(() => toPythonAstAndResolve(code, 3)).toThrow(SyntaxError);
+    // Matches CPython's own wording for this diagnostic (#187) — and, since the name reaching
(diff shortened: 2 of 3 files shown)


martin-henz:
@coderabbit review

martin-henz:
## Review: Match CPython's wording for unresolvable nonlocal declarations

### Overview

Small, well-scoped fix: `visitNonLocalStmt` (`src/resolver/resolver.ts`) now raises `ScopeConflictError` with CPython's exact wording (`no binding for nonlocal 'x' found`) instead of the generic `NameNotFoundError`, which also suppressed a misleading "Perhaps you meant to type 'x'?" suggestion. Includes tightened tests and a fix to an incidentally-affected assertion. 3 files, +25/-6.

### Correctness

- **Verified the actual rendered output** by manually triggering the error rather than trusting the diff alone:
  ```
  SyntaxError at line 4
          nonlocal x
                    ^ no binding for nonlocal 'x' found
  ```
  Matches CPython's message content and the formatting of sibling `ScopeConflictError` usages (#178-181). No off-by-one in the caret positioning.
- **`name.length` vs `stmt.name.lexeme.length`**: the diff swaps the `current` index computation to use the already-extracted `name` local. Since `name = stmt.name.lexeme` is a `const` declared at the top of the function, these are always equal — harmless simplification, not a behavior change.
- **Checked for downstream breakage**: grepped the whole codebase for anything branching on `error.name === "NameNotFoundError"` specifically (vs. just reading `.message`). Found none — `preprocess.ts` and friends only ever read `.message` generically. So collapsing this case's `.name` from `"NameNotFoundError"` to `"SyntaxError"` (via `ScopeConflictError`) is safe; nothing downstream distinguishes on it.
- **`visitGlobalStmt` scope check**: the PR description claims it checked and confirmed `global` doesn't have the analogous bug (it unconditionally accepts the declaration, deferring to a runtime check). Consistent with `visitGlobalStmt`'s code — no resolver-time "not found" path exists there to begin with, so nothing to fix. Reasonable to leave untouched, though there's no explicit regression test *proving* `global` is unaffected — minor, low-risk gap given no code changed there.

### Code quality / style

- **Good choice reusing `ScopeConflictError`** rather than introducing a new error class — its free-form `message` field and `.name = "SyntaxError"` are a closer match to CPython's actual exception shape than `NameNotFoundError`'s own invented class name, and the new message (lowercase, no trailing period) is stylistically consistent with the sibling #178-181 messages, unlike `NameNotFoundError`'s capitalized/punctuated style it replaces.
- Comments explain the *why* (the Levenshtein-suggests-itself failure mode) rather than restating the diff — matches this codebase's conventions.
- **Nitpick**: the two tightened tests in `nonlocal.test.ts` each call `toPythonAstAndResolve(code, 3)` *twice* — once for `.toThrow(/positive/)`, once for `.not.toThrow(/negative/)` — parsing and resolving the same source twice per test. No other test in this file pairs a positive and negative check on the same input like this. Cheap to tidy: capture the error once and assert both regexes against the single `.message`.

### Test coverage

- Both issue repro cases are covered and now assert on message content, not just `SyntaxError` class — a real strengthening.
- The `wasm-functions.spec.ts` fix is correct and well-explained — the test was incidentally depending on a side effect of the old generic-fallback behavior, not testing its own stated intent. Good catch, found by running the full suite rather than just the touched file.
- No test exercises `nonlocal` inside a `lambda`/comprehension edge case or multiple `nonlocal` names failing in the same function — out of scope for a message-wording fix, but worth a mental note later.

### Risks

None significant — diagnostics-only change. Both repro cases were already correctly rejected as `SyntaxError` before this PR; only the message text and error subclass change. Full suite passing (19568/19568) after the wasm test fix is strong confirmation nothing else silently depended on the old wording.

### Verdict

Solid, well-tested fix. One optional nitpick (double-invocation in the two tightened tests) — not blocking.

